Issue #165, opened 2022-10-31
Use NPM for icon provider


n8kim1:
Right now, we use icons hosted by pixiden and by font-awesome. We save them in static files and host them. This adds bloat, and also slows down page loading time by a significant amount.
We import font-awesome hosted on the cloud, as part of light-bootstrap-dashboard; this can't really be changed.
However, _we should switch out our pixiden icons for something that's hosted by npm_. This saves on code bloat, allows us to not host static files, and reduces load time. 

We already use fontawesome (see the import in index.html; note that light-bootstrap-dashboard needs it anyways)

**Use this FA, port all pixiden to fa. then delete all pixiden files, source code in `index.html`, and source code wherever else, too.**

PR #195, opened 2022-11-07
Remove pixiden icons; switch to font-awesome that we've already imported

n8kim1:
reduces our dependencies on dead code, and reduces our use of bad import patterns. Also reduces our bundled file-size by about 25%, which is pretty neat. 
**TODO: everything in the issue desc, #165**


n8kim1:
closing since we're deprecating this frontend anyways

PR #327, opened 2022-11-28
add random avatar generation to backend

### LINKED ISSUE #238: Randomly generate avatars on backend rather than frontend
Users that have not uploaded an avatar have a randomly generated avatar; it would be nice to generate this on the backend and put it in storage (using the same UUID logic as existing), rather than do this on frontend

apollo1291:
Moved the random avatar generation to the backend

Closes #238


backend/siarnaq/api/user/models.py
```diff
@@ -102,14 +106,48 @@ def get_avatar_url(self):
...
+            avatar = Image.new("RGB", imgsize)  # Create the image
+
+            leftColor = rgb1  # Color at the right
+            rightColor = rgb2  # Color at the left
+
+            for height in range(imgsize[1]):
+                for width in range(imgsize[0]):
+
+                    # Make it on a scale from 0 to 1
+                    dis = float(width) / imgsize[0]
```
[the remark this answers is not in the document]
gaurav-arya: Also, width and height don't seem like great names here as they are usually used for the full width/height of the image; maybe `x` and `y` or `pixel_x` and `pixel_y`?

gaurav-arya:
Is this ready for review yet?

backend/siarnaq/api/user/random_avatar.py
```diff
@@ -0,0 +1,70 @@
...
+    return avatar
+
+
+def get_avatar_url(entity):
+    """Return a cache-safe URL to the avatar."""
+    # To circumvent caching of old avatars, we append a UUID that changes on each
+    # update.
+
+    # check whether entity is user
+    if type(entity).__name__ == "User":
```
gaurav-arya, line 50: This isn't good because it hardcodes what sort of entities this function allows

backend/siarnaq/api/user/random_avatar.py
```diff
@@ -0,0 +1,70 @@
...
+            r = rightColor[0] * dis + leftColor[0] * (1 - dis)
+            g = rightColor[1] * dis + leftColor[1] * (1 - dis)
+            b = rightColor[2] * dis + leftColor[2] * (1 - dis)
+
+            avatar.putpixel((pixel_x, pixel_y), (int(r), int(g), int(b)))
+
+    return avatar
+
+
+def get_avatar_url(entity):
```
gaurav-arya, line 44: The main thing we should definitely do is abstract away the random avatar generation code (the `generate_avatar` function); you've tried to abstract `get_avatar_url` as well, but this seems a bit brittle (e.g. see the comment about hardcoding entities). I would say the simplest solution would be to move the `get_avatar_url` logic back to user and team (copying a bit of logic, which is ok with me for now) and then only keep the `generate_avatar` function here. 

If we really wanted to abstract this logic, it would probably make the most sense to have a custom avatar field with a `get_avatar_url` member function. But that is a bit messy since it changes e.g. the DB model, so probably out of scope for this PR. Pinging @j-mao for any thoughts

backend/siarnaq/api/user/random_avatar.py
```diff
@@ -0,0 +1,70 @@
...
+
+    # check whether entity is user
+    if type(entity).__name__ == "User":
+        avatar_path = posixpath.join("user", str(entity.pk), "avatar.png")
+    else:
+        avatar_path = posixpath.join("team", str(entity.pk), "avatar.png")
+
+    if not entity.has_avatar:
+
+        entity.has_avatar = True
```
gaurav-arya, line 57: so `has_avatar` indicates "has uploaded avatar". Which should be true only if a user has uploaded an avatar, not if they have a randomly generated one. So I think `has_avatar` should be kept `False` until a user actually uploads an avatar. (Perhaps it could be renamed `has_uploaded_avatar`...)

Instead, for randomly generating avatars, we should have `avatar_uuid` be null by default and set it to a value once a randomly generated avatar has been created (for existing users in prod, we would need to do a DB update to set `avatar_uuid` to null if `has_avatar` is False). This is a bit low priority, so I would suggest prioritizing #370 first.

gaurav-arya (changes requested):

gaurav-arya:
Is this ready for another review yet?

j-mao (changes requested):
Noting here that this needs to be rebased after #390 is merged, because the migration needs to be re-generated.

acrantel:
Closing b/c we'll display a default profile picture instead
https://github.com/battlecode/galaxy/issues/825

PR #342, opened 2022-11-29
Make divs on supporting pages responsive, so that they fit and work on small screens

n8kim1:
Things used to look like:
image

Now the div fits on the screen (including auto resizing!)
tests fine on Chrome with responsive device settings, although I'd like to verify on a proper mobile device sometime

### DIFF
--- a/frontend/src/views/login.js
+++ b/frontend/src/views/login.js
@@ -54,7 +54,7 @@ class LoginRegister extends Component {
           className="card"
           style={{
             padding: "20px",
-            width: "350px",
+            width: Math.min(window.innerWidth, 350),
             margin: "40px auto",
             marginBottom: "0px",
             fontSize: "1.1em",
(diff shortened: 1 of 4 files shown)


j-mao (approved):
The code looks sane; I trust you've tested this works.

For future, the CSS needs to be re-done: not just to make it DRY, but also because I think we're reinventing the wheel as our hands are tied by our current framework.

n8kim1:
"I think we're reinventing the wheel as our hands are tied by our current framework." 
should point out: the support pages are all handcoded and use handcoded divs?
I wonder if we could've just used standard bootstrap divs (like, eg, what the register box itself uses. col-lg-12, col-md-6, etc)

j-mao:
I think our hands were tied by the fact that we werent using bootstrap :P if we can use it, then things will be easier for us.

Our hands are also tied by the fact that we're not using Bootstrap v5…

PR #348, opened 2022-11-29
implement throttling, limit api calls to 100 per minute

### LINKED ISSUE #331: [backend] API throttling
Search for DRF Throttles

apollo1291:
limit a user from sending more than 100 api calls per minute

closes #331

### DIFF
--- a/backend/siarnaq/settings.py
+++ b/backend/siarnaq/settings.py
@@ -168,6 +168,8 @@ class Base(Configuration):
 
     # Authentication with simple JWT
     # https://django-rest-framework-simplejwt.readthedocs.io/en/latest/getting_started.html
+    # throttling
+    # https://www.django-rest-framework.org/api-guide/throttling/
 
     REST_FRAMEWORK = {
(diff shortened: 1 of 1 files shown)


j-mao:
This looks great to me! Just wanted to check with @gaurav-arya @n8kim1 : I suggested 100/min out of the blue but don't actually know what the frontend needs- can you confirm?

n8kim1:
A full load of page takes 4 API calls (sigh). Then clicking around, takes, like, 2 API calls max per button press, unless state is reloaded so that's 6 uhhh 

one click per second max?
300 calls/min? 400 calls/min? is a conservative upper bound. is that too high? @gaurav-arya et al

gaurav-arya:
I'm good with either:) worst case would be someone spam refreshing I guess, so then something liike 400 would be safer? But happy to merge with either number, so I'll leave it up to @apollo1291's judgement?

gaurav-arya (approved):

j-mao (approved):

gaurav-arya (approved):

PR #351, opened 2022-11-30
Maximum File Length

lowtorola:
Not sure how to test this implementation of maximum file length so this is a draft PR for now... would appreciate some eyes on this.


lowtorola:
Need to test! Gaurav suggested using this on avatars to test so will try this.

j-mao:
Haven't tried it myself, but looks like it works! One of the test cases generates an awfully large file upload, and that test case now fails :P

gaurav-arya:
By the way your commits still seem to not be associated with your GitHub account; might be worth fixing that and rebasing:)

lowtorola:
> By the way your commits still seem to not be associated with your GitHub account; might be worth fixing that and rebasing:)

That's odd, they seem to be associating now on my screen. Before I "fixed" it they didn't have my profile pic but now they do?

gaurav-arya:
hm, on my end your commits still don't have a profile pic (e.g. n the commits tab of this page)

lowtorola:
> hm, on my end your commits still don't have a profile pic (e.g. n the commits tab of this page)

Ah, I see. I'm going to try to fix that at the meeting Saturday :+1:

gaurav-arya:
Hmm, strange, I don't see a profile pic on your commits on my end

backend/siarnaq/api/compete/serializers.py
```diff
@@ -61,10 +63,36 @@ def update(self, instance, validated_data):
...
+
+    def __eq__(self, other):
+        return isinstance(other, FileValidator) and self.max_size == other.max_size
+
+
 class SubmissionSerializer(serializers.ModelSerializer):
     teamname = serializers.CharField(source="team.name", read_only=True)
     username = serializers.CharField(source="user.username", read_only=True)
-    source_code = serializers.FileField(write_only=True)
+    validate_file = FileValidator(max_size=5 * 1024 * 1024)
```
gaurav-arya, line 94: I think it's confusing to have this as a class variable if it's only needed for defining the line below. Happy if you just inline this into the below line, or any other alternate solution where we don't have an extra class variable

gaurav-arya (dismissed):
Looks good to me, other than the comment I left! But wait for an additional review from @j-mao

backend/siarnaq/api/compete/serializers.py
j-mao: Agree with this- otherwise all looks good to me!

lowtorola:
Looks like the change has to be approved before I can merge :rocket:

gaurav-arya (dismissed):
Done! Please make sure to give the latest version a quick test in staging, then feel free to merge :)

lowtorola:
@gaurav-arya It seems to want this PR to pass all of the checks before it will let me merge... is there a workaround for this? Kinda funny because this PR intentionally fails the "large file submission" Django test Jerry mentioned earlier.

j-mao:
Hahaha good point. I think that test case is now bad because it… isn't supposed to work anymore :P

I'll push a change to this test that makes it smaller :')

j-mao (approved):

PR #369, opened 2022-12-03
Create Public User Page

### LINKED ISSUE #19: Create a public user page
Similar to the team-info page we have -- a page with info for each user. (Not sure what might go on here / what's could be considered public)

A couple reasons why:
So that users can individually advertise their own info. (bio, links to other websites like personal website or LinkedIn or github or whatevs)
So that users can be on multiple teams across multiple competitions. Handy if we make a all-version scrimmaging website

**note that this user page should be agnostic to episode!! Feel free to ask for help/clarification about what this means, or how to implement**

lowtorola:
Resolves #19 by creating a bare-bones public user page utilizing the existing `userCard` and `teamCard` components. Draft for now, work incoming :rocket:!

### DIFF
--- a/backend/siarnaq/api/user/views.py
+++ b/backend/siarnaq/api/user/views.py
@@ -19,6 +19,8 @@ from siarnaq.api.user.serializers import (
     UserPublicSerializer,
     UserResumeSerializer,
 )
+from siarnaq.api.teams.models import Team
+from siarnaq.api.teams.serializers import TeamPublicSerializer
 from siarnaq.gcloud import titan
 
 logger = structlog.get_logger(__name__)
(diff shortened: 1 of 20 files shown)


n8kim1:
cross-checked other examples, and noticed braces instead of parens. Not 100% sure why that fixes but it fixes! I think because in braces, it was treating as code (so, like, the line just declares a component, but outputs nothing / silently drops it). 
Anyways there's new errors but you should be able to see them now in console etc

n8kim1:
Frontend looks good! However, explicitly tagging @j-mao for backend review, so I'll hold off on clicking the approve button for now
(would tag gaurav too but don't wanna give him too much on his plate during these times)

Have a couple small things to fix, but they're nowhere close to as urgent as other p1 things. So I'll file follow-up issues and we can just focus on more pressing work

backend/siarnaq/api/user/views.py
```diff
@@ -143,3 +145,31 @@ def avatar(self, request, pk=None):
...
+
+    permission_classes = (AllowAny,)
+    filter_backends = [filters.SearchFilter]
+
+    def get_queryset(self):
+        return Team.objects.all()
+
+    @action(
+        detail=False,
+        permission_classes=(IsAuthenticated,),
```
j-mao, line 165: Would `AllowAny` be enough here? I think non-logged-in people can also be allowed to see team memberships.

j-mao (changes requested):
I think this looks great! One thing I'm a bit queasy about, though, is the whole extra viewset… would it be possible to move the `teams` method into the existing `UserViewSet`? Please let me know / comment in code if there was a specific reason this isn't possible.

Perhaps this is achievable by:
- Moving the entire `teams` method over/
- Setting `detail=True`. Then, you'll receive the user ID directly as a kwarg `pk`.
- Instead of using `self.get_queryset()` in `teams`, you can use `Team.objects.all()` directly. Perhaps even filter to `Team.objects.all().visible()` depending on whether `request.user` is staff.

lowtorola:
Thanks Jerry! I struggled for at least a day trying to get the endpoint under the UserViewSet... the `pk` kwarg hint solved it!

j-mao (approved):
Looks good to me! Just pushed a small commit to lint.

PR #388, opened 2022-12-08
Enable submissions, with just the ability to submit code, and improve tournament submission deadline countdown

### LINKED ISSUE #75: **Dynamic** countdown timer in frontend for tournament submission deadline
Idea: we have a tournaments table, right? Pull the tournament with the next date (or the next 2 or 3 or whatevs, in case of ties, bc this happens). Also get their names. Load these as the API "fake response". 

This requires, in DB, files for both submission deadline and tour datetime.
Any time we update submission deadlines, we'll have to remember to change the db, too. This is still much more efficient than the current process.

Carefully handle two tournaments being next, with the same deadline! The data should be pulled properly when this happens, and the frontend should update properly, when this happens. (the frontend's tour name should at least include both tournaments. Might also have to have a verb-agreement-with-subject check, but it'd be easier if the frontend info was written in a way that didn't require such changes and checks)

### LINKED ISSUE #127: Enable submission page with barebones submission functionality

n8kim1:
closes #127 and closes #75


frontend/src/api.js
```diff
@@ -19,9 +19,9 @@ const SCRIMMAGE_STATUS = {
 // NOTE: CHANGE THESE FOR EVERY UPCOMING TOURNAMENT AND DEADLINE
 // Note the UTC timezone, change for daylight savings, etc, as necessary
 // (UTC-5 is the usual timezone for Cambridge in January)
-const HAS_NEXT_TOUR = false;
+const HAS_NEXT_TOUR = true;
 const NEXT_TOUR_SUBMISSION_DEADLINE = new Date(
-  "January 27, 2022 19:00:00-5:00"
+  "January 27, 2023 19:00:00-5:00"
```
n8kim1, line 24: must revert before merging!!!!

n8kim1:
@gaurav-arya Might be a messy review; Would suggest just looking at code additions (not deletions) since a lot of deleted stuff is legacy anyways. Also, no need to clean out code that I didn't delete yet -- saving for future PR

gaurav-arya (dismissed):
Looks good to me. Thought it would be easier to just go ahead and skip the intermediate stage and implement the next tournament API here, since it would be a pain to do it manually in IAP. Please look over changes given that they're fairly substantial:) Also requesting a review from @j-mao to look over backend endpoint + maybe give a usage test of the countdowns given that this will go straight to a public facing page in prod

backend/siarnaq/api/episodes/views.py
```diff
@@ -84,6 +86,16 @@ def get_queryset(self):
             queryset = queryset.filter(is_public=True)
         return queryset
 
+    @action(detail=False, methods=["get"])
+    def next(self, request, episode_id):
+        "Retrieve the next upcoming tournament, as ordered by submission freeze time."
+        now = timezone.now()
+        next_tournament = get_object_or_404(
+            self.get_queryset().filter(submission_freeze__gte=now)
```
j-mao, line 94: This will crash if there is >1 tournament in the future, because `get_object_or_404` does not work with querysets of length ≥ 2. Use a slice to limit the size of the returned queryset. Note that line 80 (`.order_by("submission_freeze")`) guarantees that the first one is indeed the chronologically next one, and not an arbitrary future tournament.

```suggestion
            self.get_queryset().filter(submission_freeze__gt=now)[:1]
```

j-mao (changes requested):

frontend/src/components/countdown.js
```diff
@@ -1,53 +1,65 @@
...
+        //   }
       });
-    });
+    } else {
+      this.setState({ countdown_time: new Date(this.props.game_release) });
+    }
+    this.refreshCountdown();
+    this.interval = setInterval(() => {
+      this.refreshCountdown();
+    }, 1000);
```
j-mao, line 49: This doesn't work in a live-test, I always get a blank countdown.

I think it's caused by a "race" (not really): the callback in lines 27-29 is async. So, these lines happen first, and the interval is set before the actual countdown time is known. `this.refreshCountdown` therefore calls `this.stop()` and kills the countdown before the callback runs.

I think the fix is to move these lines into the callback?

j-mao (changes requested):
Also: `frontend/src/views/submissions.js:38` doesn't match the method signature of `Api.getNextTournament` and causes this misrendering of the Submissions page.

image

j-mao (commented):
Gave it a whirl and it looks like it works! :tada:

A couple small notes:
1. Django logs still show some broken API calls:
```
GET /api/episode/function%20(tournamentInfo)%20%7B%20%20%20%20%20%20%20%20_this2.setState(%7B%20%20%20%20%20%20%20%20%20%20tournamentInfo:%20tournamentInfo%20%20%20%20%20%20%20%20%7D)%3B%20%20%20%20%20%20%7D/tournament/next
```
2. The instructions are bulky and should be slimmed down. Most of them are true but no longer important, because they are relics from when the submission page was unstable/buggy. Removing them will make the UI cleaner.

   More specifically, I think the following can be dropped:
   - "For peace of mind, ..." –> not our job to tell people how to manage their time
   - "stay on this page until" –> the submission page no longer has this failure mode
   - "if bot does not compile" plus the entirety of the "compiling tips" panel –> they receive compiler output via logs and can debug this themselves
   - Soon but not yet: "create a zip with this structure" –> we'll provide a script that does this, in the scaffold
   - You might notice that this list is suggesting the deletion of all the text on the page. This is intentional.
4. The form should get emptied out as soon as submission succeeds. Spam-clicking should be idempotent, and not create a spam of submissions.

In terms of importance: I think I'd like to see (3) before explicitly approving, and (2) is a very-nice-to-have, while (1) is a kinda-nice-to-have.

j-mao:
For what it's worth, some alternatives to (3) are:

- Alternative 1: Have separate pages for submission and history. Successfully submitting then triggers a redirect to the history page, for you to view the status of your submission -- while also making it impossible to click the submit button again.
- Alternative 2: Have an additional submission details page for each submission, to view logs, status, source code, etc. Successfully submitting then triggers a redirect to the details page for the newly created submission.

I guess: common theme here is "redirect to somewhere logical". I think, the easiest solution is "empty out the form", but in the long-term we'll want to implement some form of redirect.

backend/siarnaq/settings.py
```diff
@@ -411,8 +411,8 @@ class Production(Base):
 
     GCLOUD_BUCKET_PUBLIC = "mitbattlecode-production-public"
     GCLOUD_BUCKET_SECURE = "mitbattlecode-production-secure"
-    GCLOUD_TOPIC_COMPILE = "production-siarnaq-compile"
-    GCLOUD_TOPIC_EXECUTE = "production-siarnaq-execute"
+    GCLOUD_TOPIC_COMPILE = "projects/mitbattlecode/topics/production-siarnaq-compile"
+    GCLOUD_TOPIC_EXECUTE = "projects/mitbattlecode/topics/production-siarnaq-execute"
```
j-mao, line 415: Is ok to leave like this - although ideal solution would be to use [this method](https://cloud.google.com/python/docs/reference/pubsub/latest/google.cloud.pubsub_v1.publisher.client.Client#google_cloud_pubsub_v1_publisher_client_Client_topic_path) instead in `backend/siarnaq/api/compete/managers.py:43`. I can do this separately if you'd prefer (just make an issue and assign it to me).

Would look something like reverting these changes and then writing
```python
topic=publish_client.topic_path(settings.GCLOUD_PROJECT, self._publish_topic),
```

n8kim1: #417 (cuz prioritizng)

frontend/src/views/submissions.js
```diff
@@ -58,55 +63,36 @@ class Submissions extends Component {
...
-        clearInterval(this.interval);
+    this.setState({ upload_status: "loading" });
+    Api.newSubmission(
+      this.state.selectedFile,
+      this.state.package,
+      this.state.description,
+      this.props.episode,
+      (success) => {
+        if (success) {
+          this.setState({ upload_status: "success", selectedFile: null });
```
j-mao, line 74: Perhaps, ideally also blank out the rest of the form, in addition to just the file. That way it actually looks like you reset the form, instead of having half of it disappear.

j-mao (approved):

j-mao (approved):

Issue #363, opened 2022-12-01
Delete empty teams with no submissions


j-mao:
Empty teams need to be kept around if they have active submissions. However, if they're truly empty, we can safely delete them -- and we should, to free up some space for team names.

RudranshGoel:
Could you please tell where the files are located?

PR #390, opened 2022-12-11
Delete inactive teams with no submissions.

apollo1291:
fixes Issue #363. deletes empty teams that have no submissions once the last user leaves. Does not delete teams that already have no users.


gaurav-arya:
Should this still be a draft or marked ready for review?

gaurav-arya (changes requested):
Looks good to me other than one nit!

backend/siarnaq/api/teams/signals.py
```diff
@@ -32,11 +32,14 @@ def copy_rating_to_profile(instance, update_fields, **kwargs):
 def make_empty_team_inactive(instance, action, **kwargs):
```
apollo1291, line 32: should this function name change?

gaurav-arya: Sure, how about `handle_empty_team`?

gaurav-arya (dismissed):
Looks good to me! Let's get a review from @j-mao

backend/siarnaq/api/teams/signals.py
```diff
@@ -29,14 +29,17 @@ def copy_rating_to_profile(instance, update_fields, **kwargs):
...
-def make_empty_team_inactive(instance, action, **kwargs):
+def handle_empty_team(instance, action, **kwargs):
     if action == "post_remove":
         if instance.members.count() == 0:
-            logger.debug(
-                "team_inactive", message="Team is now inactive.", team=instance.pk
-            )
-            instance.status = TeamStatus.INACTIVE
-            instance.save(update_fields=["status"])
+            if not instance.has_active_submissions():
```
j-mao, line 35: typo
```suggestion
            if not instance.has_active_submission():
```

j-mao (changes requested):
This is almost good to go! Gave it a whirl and encountered a small bug that need fixing.

You'll likely also need to  change `TeamProfile.team` to have `on_delete=models.CASCADE` (instead of `PROTECT`). The reason is, "protect" will block deleting anything that has a profile attached, while "cascade" will delete any attached profiles. After this change, you might need to generate a new *migration* (this tells the database there's an update to the models): the command is `./manage.py makemigrations`, but feel free to give a yell in slack if something funky happens.

Once it's fixed and it's tested+working, feel free to send through another review! :')

gaurav-arya (approved):
Looks good to me! Wait for a re-review from @j-mao to merge. Also make sure to live test the latest version to ensure no silly typos:)

j-mao (approved):
Looks good to go!

Issue #375, opened 2022-12-05
Display kerberos field in user profile edit page


j-mao:
This is a new field from the backend. Looks like it got missed while porting frontend?

PR #394, opened 2022-12-14
Add kerb field to user profile edit page

apollo1291:
addresses issue #375 .


j-mao:
I don't understand how frontend works, redirecting the review.

PR #404, opened 2022-12-20
Do not impersonate service accounts on GCE

### LINKED ISSUE #403: Staging backend sporadically fails to initialise
Several incidents are generated, with [this one](https://console.cloud.google.com/errors/detail/CPiIyIfykqHpEA?project=mitbattlecode) looking like the root cause. Suspecting something to do with IAM impersonation, but not yet sure.

j-mao:
This should fix #403 in the lazy way. If this resolves the issues, then we also don't have to worry about the causes anymore because we'll no longer have such a reliance on the IAM API.

### DIFF
--- a/backend/siarnaq/settings.py
+++ b/backend/siarnaq/settings.py
@@ -22,6 +22,7 @@ import google.cloud.secretmanager as secretmanager
 import structlog
 from configurations import Configuration, values
 from google.auth import impersonated_credentials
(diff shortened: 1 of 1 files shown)


n8kim1 (approved):
wanna deploy asap just to see if things fix?

j-mao:
Ya let's do it

PR #415, opened 2022-12-26
Added loading indication for the password forget interface 

### LINKED ISSUE #22: Add loading indications to password reset interfaces
When buttons are clicked, there is no indication of activity until after things load; this makes some pages seem like they're doing nothing. Users might get frustrated and assume "site's down", or they could spamclick, both are bad

maggiehuiliyao:
closes #22

### DIFF
--- a/frontend/src/views/passwordChange.jsx
+++ b/frontend/src/views/passwordChange.jsx
@@ -8,6 +8,7 @@ class PasswordChange extends Component {
     passwordVerify: "",
     success: false,
     error: false,
+    loading: false,
   };
 
   changePassword = (e) => {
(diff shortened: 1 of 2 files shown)


maggiehuiliyao:
i'm not sure how you wanted the loading interface to work so if you want me to change it, let me know! this time I managed not to commit over a hundred files yay :>

n8kim1:
Looks good _so far_! Could you add an indicator to the `passwordChange` page as well? Should be nearly identical

n8kim1 (changes requested):
(add for passwordChange.jsx)

n8kim1:
force push to fix merge conflicts that were my cause

PR #416, opened 2022-12-28
Show submission history and fix misc submission bugs

### LINKED ISSUE #387: Enable submission table for general submission history
And remove the card that showed just the last 1 submission; meld it into this general table

Also, **rerender this table when a submission is properly created (on callback)**

n8kim1:
closes #387


j-mao (commented):
Looks like it works. A couple questions:

- How does pagination work? Perhaps I'm missing something, but it seems like the code is hardwired to fetch page 1.
- Shall we also show the submitted package name?

gaurav-arya (approved):
Agree w/ Jerry's comment. Also, a small thing, but after clicking the refresh button, the button stays dark for some reason

gaurav-arya (changes requested):
Bug from live testing:

- If I upload a submission, submit it, and then try to submit a new file without refreshing the page, the file upload does not work (it keeps saying "no file chosen").
- After uploading one submission, if I click choose file with the intention of uploading a second one, but then escape the file upload window instead, the page crashes.

j-mao (dismissed):

j-mao (approved):

PR #422, opened 2022-12-29
Enable basic scrimmage requests

gaurav-arya:
This is basic functionality for requesting scrimmages. The scrimmage inbox and scrimmage history are both hidden, and autocomplete is disabled.


n8kim1 (approved):
works fine from UI! code could use some cleanup but deal w that tmrw, or after all the scrimmage PRs are done

j-mao (dismissed):
Looks great, love the popup!

Approving for now- but would be good to either hide "self" from the list of teams to scrim against, or display the error returned by backend. Otherwise, backend rejects the scrim request and frontend displays a rather cryptic message.

Feel free to either implement, or merge+create_an_issue.

n8kim1 (dismissed):

frontend/src/components/rankingTeamList.js
```diff
@@ -31,40 +95,63 @@ class RankingTeamList extends TeamList {
...
+                    className="btn btn-xs"
+                    onClick={(event) => {
+                      event.stopPropagation();
+                      this.openRequestForm(team);
+                    }}
+                  >
+                    {buttonContent}
+                  </button>{" "}
+                </td>
+              )}
```
j-mao, line 143: Do you think it'd be possible to move the `<td>...</td>` tags outside the `canRequest` conditional? I imagine you'd want a blank cell to show up if you can't request, rather than no cell (like it is now, see screenshot)

image

gaurav-arya: `canRequest` is about whether to show the whole column in the first place, so that's ok, but we should keep the td tag when conditionally rendering dependent on if it's their own team, so have fixed that

j-mao (approved):

n8kim1 (approved):

PR #424, opened 2022-12-31
Show better errors in frontend

### LINKED ISSUE #315: Refactor the recursion/printing for nested errors
See #338 for more context. Also note that our recurser in error-handling.js just breaks on some unexpected formats, is hard to deal with, is a major source of bugs, etc etc

A better fix is a generic recurser that will recurse anything. Let's do it!

### LINKED ISSUE #338: error_handling.js can obscure error messages; dump in console
error_handling.js is good at parsing many error messages, and turning them into human-readable form.
However, sometimes the error message we get is uncommon, or the result of a deeper error. In this case, our parser will parse it...but because it's not getting expected input, it gives bad output. This can confuse devs, make for extra digging, etc.

We _could_ have a more robust handler. But this is really hard -- it's hard to balance human-readability with covering all cases. There's a reason we have a custom parser, after all.

The better idea is to simply **dump the full error message/array in the console.** (Not insecure -- since the network response in console can give it to us, and motivated ppl would just check that anyways) And it's much more convenient.

Side-note: from dev experience, you should prob learn that error_handling misparses.
**Should we include a note in readme? "If response looks weird, open your console and retry and see what's printed".**

### LINKED ISSUE #421: Need to give indication on too few characters for password reset

### LINKED ISSUE #423: Team name errors (eg ascii-restriction) are uninformative
Errors should be propagated from backend, instead of using a pre-packaged error message that is probably the mode error but doesn't cover all validation errors.

n8kim1:
Closes #315 and closes #338 and closes #423 and closes #421

That was a fun 3 hours
Better error handling in general. Feel free to test, See screenshots for examples

image

image

image

### DIFF
--- a/frontend/src/utils/error_handling.js
+++ b/frontend/src/utils/error_handling.js
@@ -24,4 +24,17 @@ function parse_errors(input) {
   return result;
 }
 
+// Copy-paste into console for convenient testing.
+// Add more test cases too, within the function, as desired.
+function test() {
+  error = {
+    profile: {
+      gender: ['"" is not a valid choice.'],
+      country: ['"" is not a valid choice.'],
+    },
+    last_name: ["This field may not be blank."],
+  };
+  return parse_errors(error);
+}
+
 export { parse_errors };
--- a/frontend/src/utils/error_handling.js
+++ b/frontend/src/utils/error_handling.js
@@ -74,6 +74,13 @@ function print_errors(xhr) {
     return `Error: ${input.detail}`;
   }
 
+  // Similar for "non_field_errors"
+  // Note that doing this check for detail, non_field_errors, and many more
+  // can get repetitive...find a DRY way for this if needed later.
+  if (input.non_field_errors && Object.keys(input).length === 1) {
+    return `Error: ${input.non_field_errors}`;
+  }
+
   const errors = parse_errors(input);
 
   if (errors.length === 1) {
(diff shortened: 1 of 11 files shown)


gaurav-arya (changes requested):
Found a couple bugs during live testing:

Text seems to overlap with "X" button, and the X button itself doesn't seem to work on my end?
image

The error handling on the register page seems to be broken: for example, if I click register with most of the fields blank, the page crashes.

frontend/src/utils/error_handling.js
```diff
@@ -1,34 +1,116 @@
...
+      .map((error) => `In ${error.field}: ${error.msg}`)
+      .join("\n")}`;
+  }
 }
 
-function get_nested_profile_errors(response_json) {
-  return get_profile_errors("profile", response_json);
+// Copy-paste into console for convenient testing.
+// Add more test cases too, within the function, as desired.
+function test() {
```
gaurav-arya, line 98: tests should probably be elsewhere, but keeping it here seems like the best option for now. but could you call this something more descriptive, like `test_error_handling`?

gaurav-arya (approved):

gaurav-arya (changes requested):
whoops sorry approved wrong pr, not ready yet

gaurav-arya (dismissed):

gaurav-arya (approved):

PR #427, opened 2023-01-01
Add single elimination tournament runner in beta

n8kim1:
You can run a tournament runner! My code quality is poor here, and there are a lot of particular quirks that make it hard for others to run a tour. But for now this exists at least, and I want to run Sprint 1.


backend/TODO_TOUR.md
```diff
@@ -0,0 +1,66 @@
...
+## Architecture Concerns
+
+- "In progress?" isn't as robust as I'd like. maybe need an enum of a few things
+
+  - uninitialized / initialized / in progress / etc
+
+- Many backend servers exist and run, hence concurrency issues etc
+- If done stupidly, backend server could be frozen in a sad while loop
+- Scheduler to auto-start tournaments?
+- Scheduler to auto-query an endpoint, to keep the tournament running?
```
j-mao, line 61: Feels somewhat overkill to use scheduler for this. Some potential alternatives:

- Act in response to game results coming in. We only ever have to enqueue a match when another match ends. Therefore, this can be entirely event-based without any background processes.
  - Pros: No worries about concurrency. Also has very minimal manual work: press a single button to run the whole tournament!
  - Cons: If something goes wrong, it could bring down the whole tournament because every match is entangled with a bunch of other matches… forcing a re-run of the whole thing.
- Run rounds at a time. Within a single round, no match depends on any other match, so they can all be started simultaneously and run in parallel. A human clicks a button to start each round.
  - Pros: minimal additional backend logic. No additional work when match results get reported.
  - Cons: some additional manual work to start each round.

I'd lean in favor of the last option. What do you think?

backend/TODO_TOUR.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Backend hits challonge map list
+- For each api match whose state (on api) is "open"...
+  - (TODO NEED TO ENSURE THAT MULTIPLE BACKENDS DONT ACT ON SAME MATCH)
+  - Create match in backend db. Associate it with the proper tour and tour round!!!
+  - Send match to be run by saturn, via pubsub
+- For each match in backend db that is associated with a relevant tour round, and which is completed:
+  - Report it to challonge via api (https://api.challonge.com/v1/documents/matches/update)
+- For each match as above that tried to run but failed cuz Saturn
+  - Ask to retry
```
j-mao, line 37: Retrying is abstracted away from Siarnaq. You will receive a report asking to set it to "Retry" state, but you won't need to do anything about it; retrying occurs automatically.

Siarnaq uses the permanent "Errored" state if there are too many retries: you should treat this as irrecoverable and demand human intervention.

backend/TODO_TOUR.md
```diff
@@ -0,0 +1,66 @@
...
+- Rinse wash repeat til tour is done
+
+Then the next step ...
+
+- Create a public bracket on challonge.
+- As matches are streamed (or pre-released), publish matches/rounds to that challonge bracket.
+  - Is easiest via some bulk buttons on admin interface. I think this needs to be impld.
```
j-mao, line 44: See `backend/siarnaq/api/episodes/models.py` / `TournamentRound.release_status`. We can publish results in bulk on a per-round basis instead of a per-match basis, which would helpfully reduce the number of button clicks for the streamer.

This also covers the round visibility question below.

backend/TODO_TOUR.md
```diff
@@ -0,0 +1,66 @@
...
+
+- Change tour things to allow tour matches to be shown in dashboard
+
+TODO I'm missing some details about round visibility on backend server
+
+## Architecture Concerns
+
+- "In progress?" isn't as robust as I'd like. maybe need an enum of a few things
+
+  - uninitialized / initialized / in progress / etc
```
j-mao, line 56: Original intent of in-progress was as a "kill switch": manually set it to False if you'd like to stop running this tournament. Depending on the actual design (see next discussion), we could probably just drop this flag entirely.

backend/TODO_TOUR.md
n8kim1: Agree that scheduler is overkill. In favor of first option. Just cuz last-option is _way_ too much manual work overall (like over many rounds, that adds up realll quick). Also because the "re-run of whole thing" isn't mitigated either way I think, but not sure.

(Perhaps we could include a check that clears a round, and re-runs matches starting from some round? Challonge API can do this robustly iirc, so we'd just have to impl for backend)

backend/TODO_TOUR.md
n8kim1: extra db field for a bigger enum might be nice? we can hash out as we implement too

backend/TODO_TOUR.md
j-mao: Second option mitigates the re-run because we can delete matches by tournament round (well, it doesn't exist yet, but we can easily provide this in admin page). Willing to accept that manual work could be a bit much (to be clear, it's probably 20 button clicks that you can multitask without paying any attention). We could though, create a schedule to click these buttons for us. Then we get the best of both worlds?

backend/TODO_TOUR.md
j-mao: Sure- let's see what we need. Although, if this field is purely for the human eye and not for the tournament runner, I'd wonder if we can get away with just nixing it.

n8kim1:
“create a schedule” aren’t we tryna avoid that nuuu
anyways i’m down for second option for now — esp if it’s simpler rn — and
can create smth better later

backend/TODO_TOUR.md
n8kim1: "create a schedule" wait aren't we trying to avoid that

n8kim1: anyways, i'm down for 2nd option just to start. can probably easily transition to 1st afterwards

backend/siarnaq/api/episodes/challonge.py
```diff
@@ -0,0 +1,95 @@
...
+                    "name": tour_name,
+                    "tournament_type": tournament_type,
+                    "private": is_private,
+                    "url": tour_url,
+                },
+            }
+        }
+    )
+
+    r = requests.post(url, headers=_headers, data=data)
```
j-mao, line 49: `post` takes a `json` kwarg so you don't have to `json.dumps` yourself

backend/siarnaq/api/episodes/challonge.py
```diff
@@ -0,0 +1,95 @@
...
+AUTH_TYPE = "v1"
+URL_BASE = "https://api.challonge.com/v2/"
+
+
+def set_api_key(api_key):
+    """Set the challonge.com api credentials to use."""
+    _headers["Authorization"] = api_key
+
+
+def create_tour(tour_url, tour_name, is_private=True, is_single_elim=True):
```
j-mao, line 30: Naming in general: please don't abbreviate. Use `tournament` instead of `tour` throughout.

backend/siarnaq/api/episodes/challonge.py
```diff
@@ -0,0 +1,95 @@
...
+            }
+        }
+    )
+
+    r = requests.post(url, headers=_headers, data=data)
+    r.raise_for_status()
+
+
+# Assumes a list of names of participants, ordered by seed,
+# better participant (ie seed #1) first.
```
j-mao, line 54: Docstring

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,74 @@ class Tournament(models.Model):
     def __str__(self):
         return self.name_short
 
-    def seed_by_scrimmage(self):
+    def get_potential_participants(self):
+        """Returns the list of participants that would be entered in this tournament,
+        if it were to start right now."""
+        # TODO incorporate real eligibility stuff
```
j-mao, line 289: See fc39b64a5e91fabc578082bf11c8e4cd4c7cd12a

backend/siarnaq/api/compete/models.py
```diff
@@ -205,14 +205,20 @@ class Match(SaturnInvocation):
...
-    """Whether players should alternate orderGbetween successive games of this match."""
+    """Whether players should alternate order between successive games of this match."""
 
     is_ranked = models.BooleanField()
     """Whether this match counts for ranked ratings."""
 
     replay = models.UUIDField(default=uuid.uuid4)
     """The replay file of this match."""
 
+    challonge_id = models.IntegerField(blank=True, null=True)
```
j-mao, line 216: Feels somewhat annoying that this needs to be stored here. I wonder if we could instead use a `through` model on the `TournamentRound` foreign key to avoid clogging this model with tournament information.

backend/siarnaq/api/episodes/models.py
```diff
@@ -259,6 +264,14 @@ class Tournament(models.Model):
     in_progress = models.BooleanField(default=False)
     """Whether the tournament is currently being run on the Saturn compute cluster."""
 
+    challonge_id_private = models.SlugField(null=True, blank=True)
+    """The Challonge ID of the associated private bracket."""
+
+    challonge_id_public = models.SlugField(null=True, blank=True)
+    """The Challonge ID of the associated private bracket."""
+
+    # TODO drop the following two fields when ready to make incompatible migrations
```
j-mao, line 273: Better now than later.

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,84 @@ class Tournament(models.Model):
...
+
+        tour_name_public = f"{self.episode.name_long} {self.name_long}"
+        tour_name_private = tour_name_public + " (private)"
+
+        # For security by obfuscation,
+        # and to allow easy regeneration of bracket
+        key = random.randint(1000, 9999)
+        # Challonge does not allow hyphens in its IDs
+        # so substitute them just in case
+        tour_id_public = (f"{self.episode.name_short}_{self.name_short}_{key}").replace(
```
j-mao, line 311: Don't need to prefix with `self.episode.name_short` because `self.name_short` should already be unique.

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,84 @@ class Tournament(models.Model):
...
+        # 1-idx seed
+        # Store team id in misc, for convenience (re-looking up is annoying)
+        # Store tour submission in misc, for consistency and convenience
+        # Note that tour submission should never change during a tournament anyways
+        # due to submission freeze. Bad things might happen if it does tho
+        participants = [
+            {"name": p.name, "seed": idx + 1, "misc": f"{p.id},{p.active_submission}"}
+            for (idx, p) in enumerate(participants)
+        ]
+        print(participants)
```
j-mao, line 329: Remove debug print.

backend/siarnaq/api/episodes/challonge.py
```diff
@@ -0,0 +1,89 @@
...
+    r = requests.put(url, headers=_headers, data=data)
+    r.raise_for_status()
+
+
+def get_tour(tour_url):
+    url = f"{URL_BASE}tournaments/{tour_url}.json"
+
+    r = requests.get(url, headers=_headers)
+    r.raise_for_status()
+    return r.content
```
j-mao, line 89: `r.json()`?

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,84 @@ class Tournament(models.Model):
...
+
+        tour = json.loads(challonge.get_tour(tour_id_private))
+        # Derive round IDs
+        # Takes some wrangling with API response format
+        # TODO move this block to challonge.py
+        rounds = set()
+        for item in tour["included"]:
+            if item["type"] == "match":
+                round_idx = item["attributes"]["round"]
+                if round_idx not in rounds:
```
j-mao, line 345: Redundant if statement: `set` already removes duplicates

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,84 @@ class Tournament(models.Model):
...
+            )
+            for round_idx in rounds
+        ]
+        print(round_objects)
+        TournamentRound.objects.bulk_create(round_objects)
+
+        self.challonge_id_private = tour_id_private
+        self.challonge_id_public = tour_id_public
+        self.in_progress = True
+        self.save()
```
j-mao, line 363: Minor optimization: use `update_fields` kwarg on save.

backend/siarnaq/api/episodes/models.py
```diff
@@ -314,6 +399,8 @@ class TournamentRound(models.Model):
     )
     """The tournament to which this round belongs."""
 
+    # TODO consider renaming this as "challonge_round_number"
```
j-mao, line 402: Better now than later.

j-mao (commented):
Some notes for now.

backend/siarnaq/api/episodes/models.py
j-mao: See #545

backend/siarnaq/api/episodes/models.py
n8kim1: jk, we're saving breaking migrations for once this PR is as ready as can be

backend/siarnaq/api/episodes/models.py
n8kim1: jk, we're saving breaking migrations for once this PR is as ready as can be

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +284,83 @@ class Tournament(models.Model):
...
+        key = random.randint(1000, 9999)
+        # Challonge does not allow hyphens in its IDs
+        # so substitute them just in case
+        tournament_id_public = f"{self.name_short}_{key}".replace("-", "_")
+        tournament_id_private = f"{tournament_id_public}_private"
+
+        # TODO support double
+        is_single_elim = True
+        participants = self.get_potential_participants()
+        # Parse into a format Challonge enjoys
```
j-mao, line 318: I kinda don't like that this model has to deal with Challonge things, outside the `challonge` module. It feels a bit coupled to me.

For example, if future devs move away from Challonge, they should be able to swap the `challonge` module out without worrying about this file.

Can this slab of code be moved into the `challonge` module? Perhaps the `bulk_add_participants` method can take a queryset as a function arg.

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +425,89 @@ class Meta:
...
+                alternate_order=True,
+                is_ranked=False,
+                challonge_id=m["id"],
+            )
+            match_participant_1_object = MatchParticipant(
+                team_id=team_id,
+                submission_id=team_submissions[team_id],
+                match=match,
+                player_index=player_index,
+            )
```
j-mao, line 497: Dead code, doesn't do anything.

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +425,89 @@ class Meta:
...
+                maps=self.maps,
+                alternate_order=True,
+                is_ranked=False,
+                challonge_id=m["id"],
+            )
+            for m in matches
+        ]
+        print(match_objects)
+        matches = Match.objects.bulk_create(match_objects)
+        matches.enqueue()
```
j-mao, line 512: Don't forget to add MatchParticipants!

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +425,89 @@ class Meta:
...
+        # kes some wrangling with API response format
+        for item in tournament["included"]:
+            if item["type"] == "match":
+                round_idx = item["attributes"]["round"]
+                if round_idx == self.challonge_id:
+                    # Only enqueue the round if all matches are ready and open.
+                    # TODO create a force-requeue,
+                    # which allows for open, but not pending
+                    if item["attributes"]["state"] in ("open", "pending"):
+                        # TODO return some sort of 400
```
j-mao, line 447: Probably 409 CONFLICT

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +425,89 @@ class Meta:
...
+        )
+        # Derive matches of this round
+        matches = []
+        # kes some wrangling with API response format
+        for item in tournament["included"]:
+            if item["type"] == "match":
+                round_idx = item["attributes"]["round"]
+                if round_idx == self.challonge_id:
+                    # Only enqueue the round if all matches are ready and open.
+                    # TODO create a force-requeue,
```
j-mao, line 444: Force-requeue is available on the Matches page. I think it's unnecessary for the tournament / round.

backend/siarnaq/api/episodes/models.py
n8kim1: yup yup plan on it -- there was a "# TODO move this block to challonge.py" comment elsewhere, and I meant to put it everywhere, but ya it gonna be future cleanup

backend/siarnaq/api/episodes/models.py
n8kim1: (for now but not for long)

backend/siarnaq/api/episodes/models.py
n8kim1: track for later

backend/siarnaq/api/episodes/models.py
n8kim1: filed in #549

backend/siarnaq/api/episodes/models.py
n8kim1: coming soon!

backend/siarnaq/api/episodes/models.py
n8kim1: wait to clarify -- this is force-requeue all matches in round, if round completely dies

n8kim1: oh wait shoot massive typo lolol

backend/siarnaq/api/compete/models.py
n8kim1: we looked into it and turns out you can't for manytomany models

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,90 @@ class Tournament(models.Model):
...
+        populate the Challonge brackets, and create TournamentRounds.
         """
-        raise NotImplementedError
+
+        tournament_name_public = f"{self.episode.name_long} {self.name_long}"
+        tournament_name_private = tournament_name_public + " (private)"
+
+        # For security by obfuscation,
+        # and to allow easy regeneration of bracket
+        key = random.randint(1000, 9999)
```
j-mao, line 309: Too easy to brute-force search. How many chars are we allowed to add? Can we add lots?

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,90 @@ class Tournament(models.Model):
...
+    def initialize(self):
         """
-        Seed the tournament with eligible teamsn in order of decreasing rating, and
-        populate the Challonge brackets.
+        Seed the tournament with eligible teams in order of decreasing rating,
+        populate the Challonge brackets, and create TournamentRounds.
         """
-        raise NotImplementedError
+
+        tournament_name_public = f"{self.episode.name_long} {self.name_long}"
```
j-mao, line 304: `episode.name_long` not necessary. See live database for examples showing that `self.name_long` would suffice.

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,90 @@ class Tournament(models.Model):
...
+        # and to allow easy regeneration of bracket
+        key = random.randint(1000, 9999)
+        # Challonge does not allow hyphens in its IDs
+        # so substitute them just in case
+        tournament_id_public = f"{self.name_short}_{key}".replace("-", "_")
+        tournament_id_private = f"{tournament_id_public}_private"
+
+        # NOTE: We don't support double elim yet.
+        # Tracked in #548. (Also make sure to actually read the "style" field)
+        is_single_elim = True
```
j-mao, line 317: Don't like this. Why can't the function just take a `TournamentStyle` parameter?

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,90 @@ class Tournament(models.Model):
...
+        # Tracked in #548
+        round_objects = [
+            TournamentRound(
+                tournament=self,
+                challonge_id=round_idx,
+                name=f"{tournament_name_private} Round {round_idx}",
+            )
+            for round_idx in rounds
+        ]
+        print(round_objects)
```
j-mao, line 361: Remove.

backend/siarnaq/api/episodes/models.py
```diff
@@ -270,12 +283,90 @@ class Tournament(models.Model):
...
+        TournamentRound.objects.bulk_create(round_objects)
+
+        self.challonge_id_private = tournament_id_private
+        self.challonge_id_public = tournament_id_public
+        self.in_progress = True
```
j-mao, line 366: We no longer need `in_progress`. Let's drop it.

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +431,96 @@ class Meta:
...
+                        # rather than 500-ing
+                        raise Exception
+                    matches.append(item)
+
+        # Map participant "objects" with IDs for easy lookup
+        participants = dict()
+        for item in tournament["included"]:
+            if item["type"] == "participant":
+                id = item["id"]
+                participants[id] = item
```
j-mao, line 466: Here, and similarly for `round_idx` above:

```suggestion
            match item:
                case {"type": "participant", "id": id}:
                    participants[id] = item
```

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +431,96 @@ class Meta:
...
+            )
+            match_objects.append(match_object)
+
+            # NOTE these var names suck. Track in #549
+            # or maybe fix earlier if anyone has ideas
+            maps_for_match_objects_batch = [
+                Match.maps.through(match_id=match_object.pk, map_id=map_id)
+                for map_id in self.maps.all()
+            ]
+            maps_for_match_objects += maps_for_match_objects_batch
```
j-mao, line 488: I mean, you don't have to construct these in this loop? Can just use a listcomp outside. Something like

```py
maps_for_match_objects = [
    Match.maps.through(...)
    for match in match_objects
    for map_id in self.maps.all()
]
```

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +431,96 @@ class Meta:
...
+                # Note that player_index is 0-indexed.
+                # This may be tricky if you optimize code in #549.
+                player_index=1,
+            )
+            match_participant_objects.append(match_participant_2_object)
+
+        # match objects is a list!
+        matches = Match.objects.bulk_create(match_objects)
+        Match.maps.through.objects.bulk_create(maps_for_match_objects)
+        MatchParticipant.objects.bulk_create(match_participant_objects)
```
j-mao, line 524: Wrap in `transaction.atomic()` decorator.

backend/siarnaq/api/episodes/models.py
```diff
@@ -338,3 +431,96 @@ class Meta:
...
+            if item["type"] == "match":
+                round_idx = item["attributes"]["round"]
+                if round_idx == self.challonge_id:
+                    # Only enqueue the round if all matches are "open".
+                    # TODO handle ability to requeue a match
+                    # TODO handle ability to requeue an entire round
+                    if item["attributes"]["state"] != "open":
+                        # TODO return some sort of 400, prob 409
+                        # rather than 500-ing
+                        raise Exception
```
j-mao, line 458: Not the models's responsibility to determine HTTP response. This function should either raise a failure exception, or return a boolean indicating success. It is up to the caller (eg. a view, or the admin interface), to handle this return and translate it into an HTTP response.

j-mao (commented):
Easy to leave TODOs saying "tracked in \<some issue\>" but forget about them in the future. Can we guarantee that:
1. The issues will actually get resolved asap.
2. The issues list the specific items they refer to. It's easy to miss one of these comments if you come back to them in the future.

n8kim1:
"Easy to leave TODOs..."
@j-mao current plan is: I'll be the one who does those other two issues anyways. I'll ctrl-f well (and the issue description also says to ctrl-f, so hopefully reviewers etc will clean up)? #548 is ofc needed in a week-and-half, #549 yall can keep me accountable.
I'm also happy to laundry list in the issue descriptions, but would pref to later (cuz dev is hard)

backend/siarnaq/settings.py
```diff
@@ -319,7 +319,7 @@ class Staging(Base):
 
     GCLOUD_SERVICE_EMAIL = GALAXY_ADMIN_EMAILS[0]
     GCLOUD_LOCATION = "us-east1"
-    GCLOUD_ENABLE_ACTIONS = True
+    GCLOUD_ENABLE_ACTIONS = False
```
n8kim1, line 322: nomerge!

backend/tour/tour-data.json
```diff
@@ -0,0 +1,536 @@
+{
```
n8kim1, line 1: nomerge!

n8kim1:
merging now. I _could_ deploy from this branch, but with breaking migrations, it's gonna be really weird if another deploy happens while this isn't on main (and thus my breaking db changes get overwritten)

PR #428, opened 2023-01-01
Show request loading spinner when registering

gaurav-arya:
Note that this is on top of #424


n8kim1:
can't test cuz on phone -- assuming you did, pls do if not -- code looks legit

n8kim1 (approved):

Issue #40, opened 2021-01-11
During submission freeze, disable submissions and scrims, and clarify why


n8kim1:
check db to see if database has submissions active. then:
add some paragraphs to display; disable buttons maybe; etc.

Slightly blocked by backend (cuz of that db check), but the check is easy to fix.

n8kim1:
closed by #431

PR #431, opened 2023-01-03
Show info about freeze for submissions and scrimmage reqs

n8kim1:
for #40

### DIFF
--- a/frontend/src/components/rankingTeamList.js
+++ b/frontend/src/components/rankingTeamList.js
@@ -152,6 +152,13 @@ class RankingTeamList extends Component {
             <div className="header">
               <h4 className="title">Rankings</h4>
             </div>
+            {this.props.episode_info.frozen && (
+              <div className="content">
+                Scrimmages may not be requested, due to a submission freeze for
+                a tournament. (If you think the freeze has passed, try
+                refreshing the page.)
+              </div>
+            )}
             <div className="content table-responsive table-full-width">
               <table className="table table-striped">
                 <thead>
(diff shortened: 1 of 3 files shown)


frontend/src/components/rankingTeamList.js
```diff
@@ -152,6 +152,13 @@ class RankingTeamList extends Component {
             <div className="header">
               <h4 className="title">Rankings</h4>
             </div>
+            {this.props.episode_info.frozen && (
+              <div className="content">
+                Scrimmages may not be requested, due to a submission freeze for
```
gaurav-arya, line 157: Do we only want to say this when `canRequest` is true (since people can't request from the rankings page anyway?)

gaurav-arya (approved):

frontend/src/components/rankingTeamList.js
```diff
@@ -122,7 +122,7 @@ class RankingTeamList extends Component {
             )}
             <td>{team.profile.auto_accept_ranked ? "Yes" : "No"}</td>
             <td>{team.profile.auto_accept_unranked ? "Yes" : "No"}</td>
-            {this.props.canRequest && (
+            {this.props.canRequest && !this.props.episode_info.frozen && (
```
gaurav-arya, line 125: oh, this check should probably have been inside the `td` rather than outside

n8kim1: wait so i messed around w the headers and nothing bad seemed to happen?
if oddities occur then we can change later

gaurav-arya: Will be resolved by #440

gaurav-arya: (It wasn't actually a problem originally to be fair, it was only bad when the td sometimes showed and sometimes didn't which wasn't the case here. But the new behaviour (disabling) should be consistent with accept/reject.)

PR #432, opened 2023-01-03
Fix Rankings Pagination Bug

### LINKED ISSUE #382: Rankings pagination buttons disable
When at the first or last pages of the rankings at [[this page](https://play.battlecode.org/bc23/rankings)](https://play.battlecode.org/bc23/rankings), the "Previous" or "Next" buttons are not actually disabled, and clicking on them brings up a spinning wheel. Not breaking as you can simply navigate back into existing pages, just a fun little bug :)

lowtorola:
Weird that jsut using `disabled` on the pagination buttons wasn't enough, must be due to the way the onClick is written. Either way, it's fixed now :+1:

Closes #382


gaurav-arya (approved):
Looks good to me!

PR #448, opened 2023-01-05
Add saturn

### LINKED ISSUE #289: Add Saturn
Port it over from the private repo, clean out all the secrets from it.

### LINKED ISSUE #312: Fix infinite loop in Saturn Pubsub client
Once upon a time, I made this commit:
https://github.com/battlecode/saturn/commit/d1943f71a68e2a6434b7f2e549842cef1098dd25

The ostensible point was that sometimes, the receiver would get/throw (not sure which one) an error code of canceled or deadline exceeded. Originally when this happened, the receiver would just return, finish its control flow, and stop doing things. But it should not suddenly shut off at this point, but instead continue to receive messages. 
I achieved this by....wrapping in an infinite loop. 

**Thanks to this loop, simply turning off a running machine became a pain. If you send a shutdown request via GCloud (eg when terraform comes to kill them), then sometimes this fails. Thus the saturn workers become very difficult to config and redeploy.** It's really hard to do manually too.

The goal: either remove the infinite loop and handle errors via some better method, or include some way to check for shutdown requests within the loop. (Perhaps move the loop? or use a different paradigm?)

Someone who knows Go better than I should do this :(((((

j-mao:
Closes #289 and closes #312.


gaurav-arya (approved):

n8kim1 (approved):
code looks legit, have not tested at all tho

j-mao:
Merging to get it deployed into staging.

PR #450, opened 2023-01-05
Saturn devops fixes

j-mao:
This PR fixes all devops issues that cause Saturn to fail catastrophically. Previous PRs were merged too early; Saturn flew too close to the sun and its wings melted.

Remainder of stuff to do… to be saved for a separate PR:
- (major) Actually run the set of requested maps. I happened to disregard the set of maps and it only ever runs on maptestsmall.
- (major) Investigate why logs aren't going into our log ingress/analysis module.
- Investigate if resizing machines will make matches run even faster.
- Test that instance pre-emption is caught and handled gracefully.
- Make logs more beautiful (double-logging from saturn, races from siarnaq, etc).


backend/siarnaq/api/compete/models.py
```diff
@@ -163,7 +163,7 @@ def for_saturn(self):
 
     def enqueue_options(self):
         """Return the options to be submitted to the compilation queue."""
-        report_url = "https://{}{}".format(
+        report_url = "http://{}{}".format(
```
gaurav-arya, line 166: Why `http` rather than `https`?

gaurav-arya (approved):

backend/siarnaq/api/compete/models.py
j-mao: Staging doesn't run HTTPS so it fails there. Production runs HTTP–>HTTPS redirect so this change shouldn't break there.

Issue #433, opened 2023-01-04
Host visualizer and show replays


gaurav-arya:
Replay links from scrimmage history page should be opened in the visualizer and play correctly.

j-mao:
Visualizer will be available at https://releases.battlecode.org/client/battlecode23/1.0.0/visualizer.html, where:

- `battlecode23` corresponds with the name of the Maven artifact, which for engine-side legacy reasons is not the episode name. We'll add this to the Siarnaq episode API.
- `1.0.0` corresponds with the release version.

j-mao:
To be clear: the only reason this link doesn't work right now is because engine hasn't got any releases yet. The code for this has already been written; the link will start working immediately as soon as engine creates a release. We'll just need to write the code to hook into this.

gaurav-arya:
OK, but since we need a backend change for this I don't think we can incorporate the correct link into #435, just leave this as a critical issue?

j-mao:
Yep! Backend change was done in #446 so this is now unblocked.

n8kim1:
one thing: make sure to dynamically get version number from backend!
it should be rattling around in episode_info
use the public one

n8kim1:
closed by #465

PR #465, opened 2023-01-07
Correct link to visualizer replays

gaurav-arya:
Finishes up #433


n8kim1 (changes requested):
the url should include a link to the specific replay file (as an encoded query), so that the visualizer knows what to open...
you can see last year's code for an example, or feel free to ask help too cuz i can dig it up

n8kim1 (approved):

PR #468, opened 2023-01-07
show accept/reject on submissions

apollo1291:
if and only if status="OK!":
then show "Accepted" or "Rejected" (based on the accepted var)

This solution works if the submission status is anything other than ok, I am not sure that it works if status is OK! because I don't know how to create a submission that is "OK!" to test it. I think the logic is right though.

### DIFF
--- a/frontend/src/components/submissionList.js
+++ b/frontend/src/components/submissionList.js
@@ -40,7 +40,6 @@ class SubmissionList extends Component {
       // This check handles the case where a new page is requested while a
       // previous page was loading.
       if (page == this.state.page) {
-        console.log(data);
         this.setState({ submissions: data.results, pageLimit, loading: false });
       }
     });
--- a/frontend/src/components/submissionList.js
+++ b/frontend/src/components/submissionList.js
@@ -57,7 +57,7 @@ class SubmissionList extends Component {
           <td>{new Date(submission.created).toLocaleString()}</td>
           <td>
             {submission.status == "OK!" &&
-              SUBMISSION_ACCEPTED[submission.accecpted.toString()]}
+              SUBMISSION_ACCEPTED[submission.accepted]}
             {submission.status != "OK!" && SUBMISSION_STATUS[submission.status]}
           </td>
           <td>{submission.description} </td>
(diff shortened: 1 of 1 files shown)


frontend/src/components/submissionList.js
```diff
@@ -35,6 +40,7 @@ class SubmissionList extends Component {
       // This check handles the case where a new page is requested while a
       // previous page was loading.
       if (page == this.state.page) {
+        console.log(data);
```
apollo1291, line 43: will delete this once tested

frontend/src/components/submissionList.js
```diff
@@ -50,7 +56,11 @@ class SubmissionList extends Component {
       return (
         <tr key={submission.id}>
           <td>{new Date(submission.created).toLocaleString()}</td>
-          <td>{SUBMISSION_STATUS[submission.status]} </td>
+          <td>
+            {submission.status == "OK!" &&
```
apollo1291, line 60: is this logic consistent / is there a better way to do this

frontend/src/components/submissionList.js
```diff
@@ -50,7 +56,11 @@ class SubmissionList extends Component {
...
+              SUBMISSION_ACCEPTED[submission.accecpted.toString()]}
+            {submission.status != "OK!" && SUBMISSION_STATUS[submission.status]}
+          </td>
```
apollo1291, line 63: is this logic correct / is there a better way to do this?

frontend/src/components/submissionList.js
```diff
@@ -50,7 +56,11 @@ class SubmissionList extends Component {
       return (
         <tr key={submission.id}>
           <td>{new Date(submission.created).toLocaleString()}</td>
-          <td>{SUBMISSION_STATUS[submission.status]} </td>
+          <td>
+            {submission.status == "OK!" &&
+              SUBMISSION_ACCEPTED[submission.accecpted.toString()]}
```
n8kim1, line 60: typo in accepted

n8kim1, line 60: also i don't think you need toString

n8kim1 (changes requested):

n8kim1 (approved):

PR #469, opened 2023-01-08
Improve Rankings Table Links

### LINKED ISSUE #443: Improve ranking table clicking behaviour
The current behaviour of clicking on a row of a ranking table -- clicking anywhere goes to the team page -- is hard to discover and also easy to accidentally trigger. We should use a better solution such as making the team name (and its members' names) clickable

lowtorola:
- Made rating and team name cells link to team page
- Team member names link to individual user pages
Closes #443 :rocket:

### DIFF
--- a/frontend/public/bc/css/bc_styles.css
+++ b/frontend/public/bc/css/bc_styles.css
@@ -174,10 +174,6 @@ ol {
   }
 }
 
-.page-item:not(.disabled) {
-  cursor: pointer;
-}
-
 /* For the countdown display in `countdown.js` and the main page */
 .Countdown {
   margin: 10px auto;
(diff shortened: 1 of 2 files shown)


gaurav-arya (changes requested):
A couple things:

- The hover icon still shows for the whole row, even though it is not clickable?
- Could we make the names themselves clickable, rather than the cells? E.g. using a `NavLink` like here: https://github.com/battlecode/galaxy/blob/e1992096efd59849283ef20aa84e02bcb636b6ac/frontend/src/views/getting-started.js#L267

gaurav-arya (approved):

gaurav-arya (approved):

PR #478, opened 2023-01-08
Generate and check for distinct maps in scrimmage requestor

gaurav-arya:


frontend/src/components/scrimmageRequestForm.js
```diff
@@ -31,12 +31,22 @@ class ScrimmageRequestForm extends Component {
     this.requestScrimmage = this.requestScrimmage.bind(this);
   }
 
+  getRandomMaps(available_maps) {
+    const possible_maps = available_maps.slice();
```
n8kim1, line 35: you're sure this deep-copies right? potential way to test would be to run this multiple times

n8kim1 (approved):
looks mostly fine, see comment but ya

n8kim1:
also consider regenerating random maps after each scrim req, for variety
(so that users don't have to refresh pg or smth)

frontend/src/components/scrimmageRequestForm.js
gaurav-arya: We don't need a deep copy since we aren't modifying the inner structs, just the array itself

gaurav-arya:
We do regenerate the maps every time the modal is closed!

PR #479, opened 2023-01-08
Tweak submissions form col widths

gaurav-arya:


n8kim1:
fineee
I'm not actually sure what's best in practice; we can wait for user data to let us know

n8kim1 (approved):

PR #480, opened 2023-01-08
Disallow empty package name on backend

gaurav-arya:
I think this is sufficient, because in some sense the backend behaviour for empty package names isn't "wrong" so long as engine accepts it.


n8kim1:
could you open an issue to investigate backend behavior etc? but ya good for now

n8kim1 (dismissed):

n8kim1:
"file an issue" well ig this works too lmao

why submissions still "compile" seems oddddd....maybe maybe file an issue for that? could be problem in saturn or
gradle?

n8kim1 (changes requested):
does this need a migration to be generated i think it does

gaurav-arya:
ack that's above my paygrade... Do i just run `makemigrations` and push?

n8kim1:
ya, make sure CI passes too ofc

n8kim1 (approved):

j-mao:
I'm not convinced this should be enforced at a database level. Future devs may choose to support additional languages where "package" is not a concept that makes sense, in which case it should be left empty. This is a Saturn-side bug.

PR #481, opened 2023-01-08
changed file size check to be on frontend

maggiehuiliyao:
Error message now displays this (the 5 MiB is hardcoded though): 
image

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -33,6 +33,10 @@ class Api {
     data.append("source_code", submission_file);
     data.append("package", package_name);
     data.append("description", description);
+    const fileSize = submission_file.size / 1024 / 1024; // in MiB
+    if (fileSize > 5) {
+      callback(data, false);
+    }
     return $.ajax({
       url: `${URL}/api/compete/${episode}/submission/`,
       type: "POST",
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -33,10 +33,6 @@ class Api {
     data.append("source_code", submission_file);
     data.append("package", package_name);
     data.append("description", description);
-    const fileSize = submission_file.size / 1024 / 1024; // in MiB
-    if (fileSize > 5) {
-      callback(data, false);
-    }
     return $.ajax({
       url: `${URL}/api/compete/${episode}/submission/`,
       type: "POST",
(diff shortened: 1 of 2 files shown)


gaurav-arya (changes requested):
Hm, so I don't think this check should happen inside the API call. If you want to catch this on frontend, we should do it before we even call the API function and display the error message there. 

Here's an example of us doing that for the scrimmage request API call: https://github.com/battlecode/galaxy/blob/a3f67b1cdd278133c3661c59352463b7c26332a4/frontend/src/components/scrimmageRequestForm.js#L72-L76

frontend/src/api.js
```diff
@@ -33,6 +33,10 @@ class Api {
     data.append("source_code", submission_file);
     data.append("package", package_name);
     data.append("description", description);
+    const fileSize = submission_file.size / 1024 / 1024; // in MiB
+    if (fileSize > 5) {
```
gaurav-arya, line 37: Let's remove this redundant check?

frontend/src/views/submissions.js
```diff
@@ -217,7 +231,7 @@ class Submissions extends Component {
         <div className="row">
           <div className="col-md-4">
             <div className="form-group">
-              <label>Package Name</label>
+              <label>Package Name (where is RobotPlayer?)</label>
```
gaurav-arya, line 234:
```suggestion
              <label>Package Name (i.e. where is RobotPlayer?)</label>
```

frontend/src/views/submissions.js
```diff
@@ -58,6 +58,22 @@ class Submissions extends Component {
...
+      let alert_message =
+        "Ensure this file size is not greater than 5 MiB. Your current file size is " +
+        Number(fileSize.toPrecision(3)) +
+        " MiB.";
+      this.setState({
+        alert_message,
+        upload_status: "failure",
+      });
+      $("#submission-table-refresh-button").click();
+      setTimeout(() => {
```
gaurav-arya, line 72: We can save some code if we don't bother with the `setTimeout` logic here: instead, just do this check before we even set the upload_status to "loading"

gaurav-arya (changes requested):
Looking good, left a few more comments!

gaurav-arya (approved):

PR #492, opened 2023-01-09
Small changes to scrimmage request team search

j-mao:
- Teams without an active submission: remove them entirely from the list.
- Disallow ranked matches against staff. It's enforced by backend, so let's enforce on frontend too. As a side-effect, matches now default to unranked.
- Show team quote (useful because Dev team contains message about "we're running examplefuncs!")

### DIFF
--- a/frontend/src/components/rankingTeamList.js
+++ b/frontend/src/components/rankingTeamList.js
@@ -85,7 +85,10 @@ class RankingTeamList extends Component {
         </div>
       );
     } else {
-      const teamRows = props.teams.map((team) => {
+      const teamsToShow = props.teams.filter((team) => {
+        return team.has_active_submission || !this.props.canRequest;
+      });
+      const teamRows = teamsToShow.map((team) => {
         let buttonContent = "Request";
         if (state.pendingRequests[team.id]) {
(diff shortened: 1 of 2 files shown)


gaurav-arya:
Can we default matches to ranked for non-staff teams? Ranked matches are more fun:)

gaurav-arya:
You could set `is_ranked` in `componentDidUpdate`: https://reactjs.org/docs/react-component.html

gaurav-arya:
To elaborate, you would check if `props.team` is not null and then set is ranked based on that. If you wanted, you could move all of the closeModal reset state logic into here for consistency, but anything works:)

gaurav-arya:
Oh, hm, I'm no longer sure if this works, one sec

j-mao:
I think the thing I added works?

gaurav-arya (approved):
Code LGTM. The conditional check for rendering the request button is now redundant, so could remove if you want

gaurav-arya:
Oops no, it's not cause want to show own team. LGTM!

j-mao:
Thoughts on removing own team from the scrim list?

gaurav-arya:
I don't like it because the scrim list also serves as a convenient rankings page for active teams

PR #493, opened 2023-01-09
Enable team and user data export

j-mao:


n8kim1:
eh, sure
i'm pretty down if you wanna push a pandas script/notebook/etc that automates things, but this is good for now, and we need in prod asap so

n8kim1 (approved):

j-mao:
Not going to push because it's questionable whether it belongs in the repo. Want to keep this repo as a collection of documented code in prod, rather than a collection of notes and scripts. In time, this feature should be directly implemented in backend.

PR #499, opened 2023-01-09
update tournament eligibility

acrantel:
us quals: entirely us college students
intl quals: entirely intl college students
hs tour: entirely hs students (us or intl)


n8kim1 (approved):

n8kim1:
serena said i could

Issue #506, opened 2023-01-09
Filtering teamlist is just broken


n8kim1:
i'll debug harder soon. but:
<img width="460" alt="image" src="https://user-images.githubusercontent.com/14008996/211417468-a2ec9059-5a5b-4f85-b9e4-8e8371347d14.png">

n8kim1:
can't repro with the old code. I hope this isn't a prod-only bug (idk, build issues??)
Code has been removed in #507 and #513 too

Issue #504, opened 2023-01-09
Filtering and pagination don't play well together


n8kim1:
You should never filter after pagination. Instead, filter beforehand. This filtering ought to be done in backend side.

Note that, eg, searchTeam in frontend (and the corresponding backend API call) does it right. 
However, when we filter teams that don't have a submission (for scrimmage purposes), we do it very poorly.
<img width="1005" alt="image" src="https://user-images.githubusercontent.com/14008996/211397879-f3e5e0c4-83c5-4007-be53-df2a1e46660e.png">

TODO fix scrim list. would need to have filtering in backend.
TODO include this bug in a list of gotcha's somewhere.

PR #507, opened 2023-01-09
Disable filtering for now

n8kim1:
robust fix for later. See #506 and #504 for those...but for now...


n8kim1:
going to bypass review-requirement. pls do review later tho.

j-mao:
Looks ok, retroactively approving.

PR #508, opened 2023-01-09
Follow redirects on POST requests

j-mao:
http -> https redirect causes subsequent requests to be incorrectly issued as GET


j-mao:
Force merging for hotfix.

PR #513, opened 2023-01-10
Properly implement filtering by active submission

### LINKED ISSUE #504: Filtering and pagination don't play well together

j-mao:
Closes #504. Unsure what this does to #506.

### DIFF
--- a/backend/siarnaq/api/teams/filters.py
+++ b/backend/siarnaq/api/teams/filters.py
@@ -24,3 +24,10 @@ class TeamOrderingFilter(filters.OrderingFilter):
         if term.startswith("-"):
             prefix, term = "-", term[1:]
         return prefix + self.FIELD_TRANSFORMS.get(term, term)
+
+
+class TeamActiveSubmissionFilter(filters.BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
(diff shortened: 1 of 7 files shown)


n8kim1 (approved):

backend/siarnaq/api/teams/filters.py
```diff
@@ -24,3 +24,10 @@ def transform(self, term):
         if term.startswith("-"):
             prefix, term = "-", term[1:]
         return prefix + self.FIELD_TRANSFORMS.get(term, term)
+
+
+class TeamActiveSubmissionFilter(filters.BaseFilterBackend):
+    def filter_queryset(self, request, queryset, view):
+        if "has_active_submission" not in request.query_params:
+            return queryset
+        return queryset.filter(submissions__accepted=True).distinct()
```
gaurav-arya, line 33: Little strange not to check true/false'ness of has active submission param in url

Issue #514, opened 2023-01-10
Saturn is only running one game at a time?


j-mao:
Why??

j-mao:
We were very excited about adding Ordering Keys to enforce that the queue is a queue.

Unfortunately, Ordering Keys also come with [affinity](https://medium.com/google-cloud/google-cloud-pub-sub-ordered-delivery-1e4181f60bc8), which is a property that destroys our parallelism. Proposing that we remove this property for now.

PR #518, opened 2023-01-10
Turn off synchronous pull

j-mao:
I don't know the effect of this on #514, but I hope this fixes it. Will test in staging.


j-mao:
Does not resolve issue.

PR #521, opened 2023-01-11
Prevent queue page numbers from exploding

gaurav-arya:


n8kim1 (approved):
25 feels a lil smol but we can always tweak later

PR #539, opened 2023-01-13
Fancy rating rendering

### LINKED ISSUE #515: Completed matches with unfinalised rating render weirdly
image

j-mao:
Closes #515

### DIFF
--- a/frontend/src/views/queue.js
+++ b/frontend/src/views/queue.js
@@ -1,5 +1,6 @@
 import React, { Component } from "react";
 import Floater from "react-floater";
+import { NavLink } from "react-router-dom";
 import Api from "../api";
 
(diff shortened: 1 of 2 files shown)


n8kim1:
@gaurav-arya  you generally have a better eye for UI/UX than I do, could you take this one?

gaurav-arya:
Hmm I kind of liked seeing old and new ratings rather than only diff?

j-mao:
Hmm, I suppose part of the point of this PR is to show the magnitude of the rating change (which is harder to immediately see from just old and new, but very easy from the diff). Any suggestions on how to do this while also keeping old/new as you prefer?

gaurav-arya (approved):
Okay, makes sense:)

PR #541, opened 2023-01-13
Adjust deploy parameters

j-mao:
More game runners. Less webserver resource.

### DIFF
--- a/deploy/galaxy/main.tf
+++ b/deploy/galaxy/main.tf
@@ -186,7 +186,7 @@ module "saturn_compile" {
   image        = var.saturn_image
 
   max_instances = var.max_compile_instances
-  min_instances = 0
+  min_instances = var.min_compile_instances
   load_ratio    = 10
 }
 
@@ -214,7 +214,7 @@ module "saturn_execute" {
   image        = var.saturn_image
 
   max_instances = var.max_execute_instances
-  min_instances = 0
+  min_instances = var.min_execute_instances
   load_ratio    = 5
 }
 
--- a/deploy/galaxy/main.tf
+++ b/deploy/galaxy/main.tf
@@ -187,7 +187,7 @@ module "saturn_compile" {
 
   max_instances = var.max_compile_instances
   min_instances = 0
-  load_ratio    = 5
+  load_ratio    = 10
 }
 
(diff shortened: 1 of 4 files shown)


deploy/siarnaq/main.tf
```diff
@@ -138,15 +138,15 @@ resource "google_cloud_run_service" "this" {
...
 
       containers {
         image = var.image
 
         resources {
           limits = {
-            cpu    = "8000m"
-            memory = "4Gi"
+            cpu    = "2000m"
+            memory = "1Gi"
```
n8kim1, line 149: could we only halve these? (feel a little unconfident about this, especially looking at current metrics)

deploy/galaxy/main.tf
```diff
@@ -215,7 +215,7 @@ module "saturn_execute" {
 
   max_instances = var.max_execute_instances
   min_instances = 0
-  load_ratio    = 10
+  load_ratio    = 5
```
n8kim1, line 218: what does load_ratio do exactly? (can't find docs...)
and in particular are you sure you want to increase one and decrease the other?

n8kim1 (commented):
(see comments)

deploy/siarnaq/main.tf
j-mao: These graphs seem to be pretty far away from 25%. In addition, we're also reducing the number of concurrent requests per machine, which further alleviates.

image
image

deploy/galaxy/main.tf
j-mao: Not in docs because it's defined by us :')

We define it to be the number of jobs in queue per machine. This change halves the number of compile machines while doubling the number of execute machines.

deploy/siarnaq/main.tf
n8kim1: oh i think i misread graphs
i read the graphs as -- eg for bottom one -- "container utilized >=99% of memory, 11.99% of time"

but that would imply it spends more time at 99% than at 95% 
also that's, like,unintuitive and statistician-esque
I'll just trust you for now

n8kim1 (approved):

PR #558, opened 2023-01-15
add compete data to teaminfo

apollo1291:
draft: My staging is messed up so Ihave not tested. Because I cant test, I'm not refactoring the api file, more focused on functionality for the deadline. This pr changes teaminfo, just from comments. Should I add win-loss to the teamcard component?

### DIFF
--- a/backend/siarnaq/api/teams/views.py
+++ b/backend/siarnaq/api/teams/views.py
@@ -2,7 +2,7 @@ import uuid
 
 import structlog
 from django.db import transaction
-from django.db.models import F, Max, Q
+from django.db.models import F, OuterRef, Subquery
 from django.shortcuts import get_object_or_404
 from drf_spectacular.utils import extend_schema
 from rest_framework import filters, mixins, status, viewsets
@@ -10,7 +10,7 @@ from rest_framework.decorators import action
 from rest_framework.permissions import IsAuthenticated
 from rest_framework.response import Response
 
-from siarnaq.api.compete.models import Match
+from siarnaq.api.compete.models import Match, MatchParticipant
 from siarnaq.api.episodes.permissions import IsEpisodeAvailable
 from siarnaq.api.teams.exceptions import TeamMaxSizeExceeded
 from siarnaq.api.teams.filters import TeamActiveSubmissionFilter, TeamOrderingFilter
@@ -187,11 +187,21 @@ class TeamViewSet(
         team = request.data["id"]
         matches = Match.objects.filter(participants__team=team, status="OK!")
 
-        matches = matches.annotate(
-            score_1=Max("participants__score", filter=Q(participants__team=team)),
-            score_2=Max("participants__score", exclude=Q(participants__team=team)),
+        matches = Match.objects.annotate(
+            my_score=Subquery(
+                MatchParticipant.objects.filter(match=OuterRef("pk"))
+                .filter(team=team)
+                .values("score")
+            ),
+            opponent_score=Subquery(
+                MatchParticipant.objects.filter(match=OuterRef("pk"))
+                .exclude(team=team)
+                .values("score")[:1]
+            ),
+        ).filter(
+            my_score__isnull=False,
         )
-        win_count = matches.filter(score_1__gt=F("score_2")).count()
-        loss_count = matches.filter(score_1__lt=F("score_2")).count()
+        win_count = matches.filter(my_score__gt=F("opponent_score")).count()
+        loss_count = matches.filter(my_score__lt=F("opponent_score")).count()
 
         return Response({"wins": win_count, "losses": loss_count})
--- a/backend/siarnaq/api/teams/views.py
+++ b/backend/siarnaq/api/teams/views.py
@@ -185,9 +185,6 @@ class TeamViewSet(
     def record(self, request, pk=None, *, episode_id):
         """Retrieve the win/loss record of a team"""
         team = request.data["id"]
-        matches = Match.objects.filter(tournament_round__isnull=True).filter(
-            participants__team=team, status="OK!"
-        )
 
         matches = Match.objects.annotate(
             my_score=Subquery(
@@ -200,9 +197,7 @@ class TeamViewSet(
                 .exclude(team=team)
                 .values("score")[:1]
             ),
-        ).filter(
-            my_score__isnull=False,
-        )
+        ).filter(my_score__isnull=False, tournament_round__isnull=True, status="OK!")
         win_count = matches.filter(my_score__gt=F("opponent_score")).count()
         loss_count = matches.filter(my_score__lt=F("opponent_score")).count()
 
--- a/backend/siarnaq/api/teams/views.py
+++ b/backend/siarnaq/api/teams/views.py
@@ -185,7 +185,9 @@ class TeamViewSet(
     def record(self, request, pk=None, *, episode_id):
         """Retrieve the win/loss record of a team"""
         team = request.data["id"]
-        matches = Match.objects.filter(participants__team=team, status="OK!")
+        matches = Match.objects.filter(tournament_round__isnull=True).filter(
+            participants__team=team, status="OK!"
+        )
 
         matches = Match.objects.annotate(
             my_score=Subquery(
--- a/backend/siarnaq/api/teams/views.py
+++ b/backend/siarnaq/api/teams/views.py
@@ -185,15 +185,13 @@ class TeamViewSet(
     def record(self, request, pk=None, *, episode_id):
         """Retrieve the win/loss record of a team"""
         team = request.data["id"]
-        matches = Match.objects.filter(participants__team=team)
+        matches = Match.objects.filter(participants__team=team, status="OK!")
 
-        matches.annotate(
+        matches = matches.annotate(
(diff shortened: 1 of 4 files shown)


n8kim1:
Rank and PerfCard wouldn't be ready, cuz backend doesn't exist yet for that :// Backend doesn't (yet) expose rank, totals wins, or total losses. We should work on that first! (You could, thought you'd have to set up first...)
Showing rating would be doable now, although seems more efficient to do later too.

So in all, if you could get backend set up, then maybe work on the backend side first? Otherwise just hold off on this and find another cool frontend issue ://

gaurav-arya:
I believe we'd previously decided that total wins and total losses should be computed on frontend based on team history, in which case backend should be already ready. @j-mao any input?

n8kim1:
idts? after some trial,  we decided to just put this on backend. pagination
is annoying and sending all data is too big

backend/siarnaq/api/teams/views.py
```diff
@@ -173,3 +174,35 @@ def avatar(self, request, pk=None, *, episode_id):
...
 
         return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+    @action(
+        detail=False,
+        methods=["post"],
+        serializer_class=None,
+        permission_classes=(IsAuthenticated, IsEpisodeAvailable),
+    )
+    def record(self, request, pk=None, *, episode_id):
```
apollo1291, line 184: this is wildly inefficient. It's going to recaluculate on every query, but I didn't want to mess with the db

backend/siarnaq/api/teams/views.py
```diff
@@ -173,3 +174,35 @@ def avatar(self, request, pk=None, *, episode_id):
...
+        """Retrieve the win/loss record of a team"""
+        match_participations = MatchParticipant.objects.filter(
+            team_id=request.data["id"]
+        )
```
apollo1291, line 188: this filter is not functioning as i expect. The goal is to grab every Match particpant obj that is related to the team id, but it never seems to match.

gaurav-arya: Here's an example of filtering matches for a particular team: https://github.com/battlecode/galaxy/blob/0e8754516fcb667cf7f2fee92864d9f15e14d7a8/backend/siarnaq/api/compete/views.py#L304

backend/siarnaq/api/teams/views.py
gaurav-arya: I think that's probably fine, maybe with some `prefetch_related`'s. At least don't worry about it for now.

gaurav-arya: Oh, I see now that you're manually looping over the queryset with a python `for`. Yes, we should avoid that. You could use the `.count()` member function of a queryset to get its length (https://docs.djangoproject.com/en/4.1/topics/db/aggregation/), and do this for two querysets, one filtered for wins and another filter for losses.

backend/mitbattlecode-496f64277523.json
```diff
@@ -0,0 +1,12 @@
+{
+  "type": "service_account",
```
n8kim1: noting that we should rebase, to drop this entirely from commit history

apollo1291: yeah my bad one this one i didn't realize it was in my staged changes

j-mao: Is good that this key was already revoked a while ago.

backend/siarnaq/api/teams/views.py
```diff
@@ -173,3 +175,25 @@ def avatar(self, request, pk=None, *, episode_id):
...
+        """Retrieve the win/loss record of a team"""
+        team = request.data["id"]
+        matches = Match.objects.filter(participants__team=team)
+
+        matches.annotate(
+            score_1=Max("participants__score", filter=Q(participants__team=team)),
+            score_2=Max("participants__score", filter=~Q(participants__team=team)),
+        )
+
+        win_count = matches.filter(score_1__gt=F("score_2")).count()
```
apollo1291, line 195: I get an error that claims score_1 cant be resolved to a field. But I'm using annotate and filter almost copy paste from django documentation? [https://docs.djangoproject.com/en/4.1/topics/db/aggregation/](url)

j-mao: `.annotate` returns a new queryset so you need `matches = matches.annotate(...)`

backend/siarnaq/api/teams/views.py
```diff
@@ -173,3 +175,25 @@ def avatar(self, request, pk=None, *, episode_id):
...
+        permission_classes=(IsAuthenticated, IsEpisodeAvailable),
+    )
+    def record(self, request, pk=None, *, episode_id):
+        """Retrieve the win/loss record of a team"""
+        team = request.data["id"]
+        matches = Match.objects.filter(participants__team=team)
+
+        matches.annotate(
+            score_1=Max("participants__score", filter=Q(participants__team=team)),
+            score_2=Max("participants__score", filter=~Q(participants__team=team)),
```
apollo1291, line 192: score_1 works properly but score_2 returns None always even when its not suppose too.

gaurav-arya: Have you followed Jerry's suggestion regarding using `matches = matches.annotate(...)`?

apollo1291: Yeah, I just haven't pushed yet.

gaurav-arya: Does score_2 still not work?

frontend/src/views/teamInfo.js
```diff
@@ -34,10 +28,10 @@ class RankCard extends Component {
           </div>
           <br></br>
           <p style={{ textAlign: "center" }}>
-            Score:{" "}
-            {this.props.team.score === -1000000
+            Rating:{" "}
+            {this.state.rating === -1000000
               ? "N/A"
-              : Math.round(this.props.team.score)}
+              : Math.round(this.state.rating)}
```
n8kim1, line 34: Let's not "show" ranking, and only show rating? Eg:
image


(No i don't know the code for this, I faked it with inspect element)

frontend/src/views/teamInfo.js
```diff
@@ -52,17 +46,16 @@ class WinsCard extends Component {
...
         <div className="content">
           <div className="col-2-row">
             <div className="row-items-box items-box-center items-box-skinny">
-              <label>wins</label>
+              <label>Wins:</label>
               <h1>{this.props.wins}</h1>
             </div>
             <div className="row-items-box items-box-center items-box-skinny">
-              <label>losses</label>
+              <label>Losses:</label>
```
n8kim1, line 58: Drop the colons? More consistent with rest of page
(or if you want, you can put colons everywhere. higher effort tho)

backend/siarnaq/api/teams/views.py
```diff
@@ -173,3 +175,33 @@ def avatar(self, request, pk=None, *, episode_id):
...
+    @action(
+        detail=False,
+        methods=["post"],
+        serializer_class=None,
+        permission_classes=(IsAuthenticated, IsEpisodeAvailable),
+    )
+    def record(self, request, pk=None, *, episode_id):
+        """Retrieve the win/loss record of a team"""
+        team = request.data["id"]
+        matches = Match.objects.filter(participants__team=team, status="OK!")
```
n8kim1, line 188: !! Please hide tournament matches from this query!
(cuz then someone could count the difference btwn tour matches and regular matches, and use that to spoil results completely)

n8kim1 (changes requested):
looks great; just a few small changes. but base of code is good

backend/siarnaq/api/teams/views.py
n8kim1: As a hint from Jerry: "all team statistics should have .filter(tournament_round__isnull=True) as the first call"

n8kim1 (approved):

j-mao (changes requested):
Why the new route in urls?

backend/siarnaq/api/teams/views.py
```diff
@@ -173,3 +175,30 @@ def avatar(self, request, pk=None, *, episode_id):
             titan.upload_image(avatar, profile.get_avatar_path())
 
         return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+    @action(
+        detail=False,
+        methods=["post"],
+        serializer_class=None,
+        permission_classes=(IsAuthenticated, IsEpisodeAvailable),
```
j-mao, line 183: I might suggest removing IsAuthenticated, to allow this to be viewable by the public

backend/siarnaq/api/teams/views.py
```diff
@@ -173,3 +175,30 @@ def avatar(self, request, pk=None, *, episode_id):
             titan.upload_image(avatar, profile.get_avatar_path())
 
         return Response(None, status=status.HTTP_204_NO_CONTENT)
+
+    @action(
+        detail=False,
+        methods=["post"],
```
j-mao, line 181: This should probably be `get`, since it’s a non mutator.

j-mao: You need detail = True to get pk

PR #560, opened 2023-01-15
Fix deleting empty teams

### LINKED ISSUE #559: Can't leave a one-man team with (failed/inactive) submissions
In #390 we had an oversight -- if your team has _any_ submissions but no successful / active ones, then your team is attempted to be deleted. But this fails since the submissions are linked to the team and still need that link.

Probably best to also mark such a team as inactive rather than deleted. Alternatively, you could delete all the linked submissions.

Critical cuz if this happens you _actually_ can't leave a team and have to create a new account

j-mao:
Closes #559.


n8kim1 (approved):
assumign you tested, sure
(sorry can't test rn bandwidth ree)

PR #561, opened 2023-01-15
Historical Elo Endpoint

### LINKED ISSUE #371: Create endpoint that returns an arbitrary team's timestamp-vs-elo-ranking
For display in frontend's team performance charts.

Compile the data in backend. Return some mapping of timestamps-to-elo. The most straightforward one is, eg, for each scrimmage, return the scrimmage timestamp and the team's rating after. (Try to return just this data to cut down on size!)

lowtorola:
Resolves #371 :rocket:

### DIFF
--- a/backend/siarnaq/api/compete/views.py
+++ b/backend/siarnaq/api/compete/views.py
@@ -321,10 +321,13 @@ class MatchViewSet(
             OpenApiParameter(
                 name="team_id",
                 type=int,
-                description="A team to filter for.",
+                description="A team to filter for. Defaults to your own team.",
             ),
         ],
-        responses={status.HTTP_200_OK: HistoricalRatingSerializer(many=True)},
+        responses={
+            status.HTTP_204_NO_CONTENT: "No ranked matches found.",
+            status.HTTP_200_OK: HistoricalRatingSerializer(many=True),
+        },
     )
     @action(
         detail=False,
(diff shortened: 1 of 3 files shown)


lowtorola:
Looking for some input (am I even on the right track??)

backend/siarnaq/api/teams/models.py
```diff
@@ -239,3 +239,18 @@ def get_avatar_url(self):
         )
         # Append UUID to public URL to prevent caching on avatar update
         return f"{public_url}?{self.avatar_uuid}"
+
+
+class HistoricalRating(models.Model):
```
gaurav-arya, line 244: I don't see this model used anywhere: did you forget to remove it?

backend/siarnaq/api/compete/views.py
```diff
@@ -315,6 +316,59 @@ def scrimmage(self, request, pk=None, *, episode_id):
...
+                "timestamp": match.created,
+                "rating": match.participants.get(team=team_id).rating
+                if team_id is not None
+                else match.participants.get(team__members__pk=request.user.pk).rating,
+            }
+            for match in matches
+        ]
+
+        results = HistoricalRatingSerializer(
+            self.paginate_queryset(ordered), many=True
```
gaurav-arya, line 368: I don't think we need to paginate here

gaurav-arya (commented):
Looking good! Left a couple comments

backend/siarnaq/api/compete/views.py
```diff
@@ -315,6 +316,59 @@ def scrimmage(self, request, pk=None, *, episode_id):
...
+            status.HTTP_204_NO_CONTENT: "No ranked matches found.",
+            status.HTTP_200_OK: HistoricalRatingSerializer(many=True),
+        },
+    )
+    @action(
+        detail=False,
+        methods=["get"],
+        permission_classes=(IsEpisodeMutable,),
+    )
+    def historical_ranking(self, request, pk=None, *, episode_id):
```
gaurav-arya, line 337: ranking -> rating for consistency?

backend/siarnaq/api/compete/views.py
```diff
@@ -315,6 +316,59 @@ def scrimmage(self, request, pk=None, *, episode_id):
...
+        """List the historical rating of a team."""
+        queryset = Match.objects.all().filter(tournament_round__isnull=True)
+
+        team_id = parse_int(self.request.query_params.get("team_id"))
+        if team_id is not None:
+            queryset = queryset.filter(participants__team=team_id)
+        else:
+            queryset = queryset.filter(participants__team__members=request.user.pk)
+        has_invisible = self.get_queryset().filter(
```
gaurav-arya, line 346: should we show invisible teams to staff, like in the scrimmage endpoint?

backend/siarnaq/api/compete/views.py
n8kim1: gaurav's right, this in fact _needs_ to not be paginated -- this was the core problem that turned this into an issue+PR in the first place

backend/siarnaq/api/compete/views.py
n8kim1: ++
in fact _please_ call it rating! in other places we use "ranking" for an ordinal ranking (eg 1st 2nd 3rd) so this could get really confusing to devs and users otherwise

backend/siarnaq/api/compete/views.py
n8kim1: IMO no need to bother with showing them to staff, for now?

Invisible teams don't really have a _good_ notion of rating anyways (since regular people can't really scrimmage them, they don't go into autoscrims, etc). Not worth the extra effort to put the code in. (unless there's a usecase I'm missing)

lowtorola:
Fixed!

n8kim1 (approved):

backend/siarnaq/api/compete/views.py
```diff
@@ -315,6 +316,59 @@ def scrimmage(self, request, pk=None, *, episode_id):
...
+            participants__team__status=TeamStatus.INVISIBLE
+        )
+        queryset = queryset.exclude(pk__in=Subquery(has_invisible.values("pk")))
+        queryset = queryset.filter(is_ranked=True)
+
+        if queryset.all().count() == 0:
+            return Response(None, status=status.HTTP_204_NO_CONTENT)
```
j-mao, line 355: Would it make more sense to return an empty list instead? I feel like we don't need this special case, and leaving it out will have reasonable behaviour.

lowtorola:
Need new review after making changes per @j-mao request :rocket:

j-mao (approved):

PR #562, opened 2023-01-16
frontend avatar and resume file size check

### LINKED ISSUE #335: Validate sizes of resumes and avatars before uploading

maggiehuiliyao:
closes #335

### DIFF
--- a/frontend/src/components/avatarUpload.js
+++ b/frontend/src/components/avatarUpload.js
@@ -28,10 +28,10 @@ class AvatarUpload extends Component {
       } else {
         this.setState({ status: "failure" });
       }
-      setTimeout(() => {
-        this.setState({ status: "waiting" });
-      }, 2000);
     });
+    setTimeout(() => {
+      this.setState({ status: "waiting" });
+    }, 2000);
   }
 
   render() {
--- a/frontend/src/views/account.js
+++ b/frontend/src/views/account.js
@@ -102,6 +102,16 @@ class Account extends Component {
   }
 
   uploadResume = () => {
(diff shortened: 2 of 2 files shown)


maggiehuiliyao:
I'm not sure if the limit is still 5 MiB so let me know if I should change that. I also wasn't sure how to figure out how to edit the status of the avatar upload button to return back to normal so my solution was moving the timeout stuff if that makes sense. So it works supposedly but just it will show the loading symbol even though it isn't really doing anything.

n8kim1:
(putting "frontend" in title just to point that out explicitly -- we don't have backend check yet anyways oops)

n8kim1:
(will push off reviewing this til weds or smth -- i gotta focus on tour stuff)

n8kim1:
(will push off reveiw til after first-tour i think)

frontend/src/components/avatarUpload.js
```diff
@@ -28,10 +28,10 @@ class AvatarUpload extends Component {
...
         this.setState({ status: "failure" });
       }
-      setTimeout(() => {
-        this.setState({ status: "waiting" });
-      }, 2000);
     });
+    setTimeout(() => {
+      this.setState({ status: "waiting" });
+    }, 2000);
   }
```
n8kim1, line 35: could you revert this change? realizing it'll make legit things break. I'll think abt what else you could do to mitigate...

n8kim1:
looks good! might be worth refactoring later but not necessary now.

n8kim1 (approved):

PR #563, opened 2023-01-16
Disabled auto ranked scrimmages?

maggiehuiliyao:
Still attempting to fix something. Questionably works but I'm not sure if I tested it correctly?


j-mao:
Superseded by #601.

PR #564, opened 2023-01-16
Disable auto-accept for ranked scrimmages (and do nothing else)

n8kim1:
As discussed here's the quick braindead fix for now (cuz time press). 

UI/UX is probably now weird. Should announce in discord (and make frontend more intuitive later, like in Maggie's PR?)


gaurav-arya (dismissed):

j-mao:
Approved per slack, but please fix your unit tests

n8kim1:
ah shiz that’s what i get for ”fixing” without running
will do after din

j-mao (approved):

PR #566, opened 2023-01-16
Show list of tournament matches

### LINKED ISSUE #15: Easy way to view tournament match lists
Currently, teams can't see tour matches in their list of matches (since we hide them for spoilers). Would be good to find a way that hides them only temporarily.

Also, it would be good to make a staff-only page, listing all the tournament matches (and associated info) for easy access.

gaurav-arya:
Closes #15. Have made #567 for future improvements.

### DIFF
--- a/frontend/src/api.js
+++ b/frontend/src/api.js
@@ -526,11 +526,15 @@ class Api {
       });
   }
 
-  static getAllMatches(episode, callback, page) {
+  static getAllMatches(episode, tournament_id, callback, page) {
     const query_data = {
       page,
+      tournament_id,
     };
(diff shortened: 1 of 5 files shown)


n8kim1:
can test later. ps pls rebase this on main (or i can?) cuz easier to test

j-mao:
I'd prefer a global page that can view everyone's tournament matches. I think we can keep this my-team-only page, but I'd also want in addition a "all tournament matches" page.

j-mao:
I've implemented this.

n8kim1 (approved):

PR #582, opened 2023-01-17
Show replay link in tournament admin

j-mao:


n8kim1 (approved):
pls trst if haven't alrrafy

PR #591, opened 2023-01-19
Log with higher-precision timestamp

j-mao:


n8kim1 (approved):
i see now why andy calls our must-review policy a bunch of bloat and red tape and headache

PR #601, opened 2023-01-23
Implement new rules for ranked matches

### LINKED ISSUE #565: Figure out what to do about ranked scrimmages, and clean up / create code
(perhaps involving backend and frontend both)

In particular, don't skip the currently-skipped unit test! Instead fix the behavior to match what's expected.

See #563 for inspiration (or just use it)

j-mao:
We're turning them back on! But with the following new rules:

- Ranked matches must use alternating color
- Ranked matches must accept random map choice assigned by the server
- You may only have a limited quantity of outstanding ranked matches or scrimmage requests at a time, counted at a per-team-pair level.

Closes #565 and renders parts of #563 moot.


acrantel (dismissed):
lgtm

acrantel (dismissed):

acrantel (approved):

Issue #574, opened 2023-01-17
Publish tournament results to public bracket and frontend


n8kim1:
- Create a bracket on challonge. Make it "public" (which means that it appears on our user page and you can google it)
- As matches are streamed (or pre-released), publish matches/rounds to that challonge bracket.
  - Perhaps the best way to do this is thru a signal, that notes when a TournamentRound switches its visibility. **TODO how can you do this only on changes _from_ not visible to visible? If you fire an API call on any save, even when retaining visibility, then this is a useless API call and will seriously start burning thru the API limits**

n8kim1:
done in #602 and #606!!!

PR #602, opened 2023-01-24
Add public external IDs to tournament and matches

n8kim1:
When a tournament is run, this will create external IDs for all the public stuff. This should be enough to run a tournament and be able to publish things later.

If the model looks good I'll make migration then, don't wanna make-rollback-remake

Does half of #574


backend/siarnaq/api/episodes/models.py
```diff
@@ -302,10 +302,17 @@ def initialize(self):
...
-
         bracket.bulk_add_teams(self, teams, is_private=True)
+
         bracket.start_tournament(self, is_private=True)
 
+        # Also create a bracket for public display
+        # Make sure teams stay the same for consistency!
+        bracket.create_tournament(self, is_private=False)
+        bracket.bulk_add_teams(self, teams, is_private=False)
+        bracket.start_tournament(self, is_private=False)
```
j-mao, line 313: Can this just be

```python
teams = ...

for is_private in [False, True]:
    bracket.create_tournament(self, is_private=is_private)
    bracket.bulk_add_teams(self, teams, is_private=is_private)
    bracket.start_tournament(self, is_private=is_private)
```

backend/siarnaq/bracket/challonge.py
```diff
@@ -141,17 +141,63 @@ def get_round_indexes(tournament: Tournament, *, is_private: bool):
...
+    # Compute a correspondence between the two.
+    # Currently, we use the fact that IDs are in the same order for all tournaments.
+    # You could also use other fields, such as Challonge's `identifier` or
+    # `suggested_play_order` for matches; these are equal across brackets.
+    # Similarly you could use `seed` for teams.
+    matches_private.sort(key=lambda i: i["id"])
+    matches_public.sort(key=lambda i: i["id"])
+    matches_private_to_public = dict()
+    for c1, c2 in zip(matches_private, matches_public):
+        matches_private_to_public[c1] = c2
```
j-mao: `c1` is a dict?

backend/siarnaq/bracket/challonge.py
```diff
@@ -141,17 +141,63 @@ def get_round_indexes(tournament: Tournament, *, is_private: bool):
     return round_indexes
 
 
-def get_match_and_participant_objects_for_round(
-    round: TournamentRound, *, is_private: bool
-):
-    tournament_data = get_tournament_data(round.tournament, is_private=is_private)
-    # Derive match dicts/JSON objects (that Challonge gives us) of this round
+def _pair_public_private_challonge_ids(tournament: Tournament):
```
j-mao, line 144: I feel like this function does the same thing twice: once on items with type `match` and once with `participant`… and the behaviour is identical for both cases.

Perhaps accept the type as a function argument?

backend/siarnaq/bracket/challonge.py
```diff
@@ -198,6 +255,9 @@ def get_match_and_participant_objects_for_round(
             challonge_participant_id = challonge_match["relationships"][
```
j-mao, line 259: Rename to include `private` in name.

j-mao (commented):
lgtm, hope it works in staging

backend/siarnaq/api/episodes/models.py
```diff
@@ -298,7 +298,7 @@ def initialize(self):
         self.external_id_private = f"private_{self.external_id_public}_{key}"
 
         teams = self.get_eligible_teams()
-        for is_private in [False, True]:
+        for is_private in [True, False]:
```
j-mao, line 301: Just wondering, was there a reason for swapping them around?

j-mao (approved):
Looks good to me

backend/siarnaq/api/episodes/models.py
n8kim1: the overall tour workflow is basically "do private first then public!". and same for code.
When I was testing, i was looking at IDs, seeing private IDs be _higher_ than public ones. This goes against "private first then public" and so it gave me a headache and i didn't like it

PR #603, opened 2023-01-24
Allow deletion of tournament rounds

j-mao:


gaurav-arya (approved):

n8kim1:
wait eek does this not delete matches? 
i think matches _must_ be deleted if tournamentRound is deleted too for security

j-mao:
it tries to
it gets blocked because matches are protected via matchparticipants

PR #604, opened 2023-01-24
Allow matches to be deleted

j-mao:


n8kim1:
for admin panel, if you select a set of last N matches, does it just work? or do you have to select in order?

either way this should be fine yay

n8kim1:
also is it possible for tournamentround deletion to automagically delete its matches too?

n8kim1 (dismissed):
pls see comments but ya

j-mao:
> also is it possible for tournamentround deletion to automagically delete its matches too?

this was #603.

n8kim1 (approved):

PR #609, opened 2023-01-25
Force happens-before with match result and rating

j-mao:
Race condition, begone


n8kim1:
the answer is always some builtin you've never heard of til you need it and google

n8kim1 (approved):

PR #610, opened 2023-01-25
Enable double elimination tournaments

### LINKED ISSUE #548: Enable double elimination tournaments in runner
Perhaps the trickiest part is mangling rounds -- challonge's round ID is odd, and then we have to convert that to round number for human display. Also, play order for rounds is nontrivial.

Probably want to have matches played in challonge's suggested play order, for convenience.

**Make sure to ctrl-f "#548" and resolve all things in this issue (and associated PR)!!!!**

n8kim1:
Closes #548


backend/siarnaq/bracket/challonge.py
```diff
@@ -126,21 +126,118 @@ def get_tournament_data(tournament: Tournament, *, is_private: bool):
...
+            matches, key=lambda match: match["attributes"]["suggestedPlayOrder"]
+        )
+        # Give it its own round
+        match_last["attributes"]["round"] += 1
+
+    return data
 
-def get_round_indexes(tournament: Tournament, *, is_private: bool):
+
+# TODO should this be private-only, and thus should I drop the is_private flag?
```
n8kim1, line 149: @j-mao opinions?
Should we always get private data (for this method), thanks to the assumption that private and public have same structure etc?

gosh that seems odd

backend/siarnaq/bracket/challonge.py
```diff
@@ -126,21 +126,118 @@ def get_tournament_data(tournament: Tournament, *, is_private: bool):
...
+
+    for match in matches:
+        round_index = match["attributes"]["round"]
+        if round_index not in round_indexes:
+            round_indexes.append(round_index)
 
     return round_indexes
 
 
+def _get_round_indexes_and_names(tournament: Tournament, *, is_private=bool):
```
n8kim1, line 171: always default to private? is that odd

backend/siarnaq/bracket/challonge.py
```diff
@@ -126,21 +126,118 @@ def get_tournament_data(tournament: Tournament, *, is_private: bool):
...
+        round_indexes_and_names.append(
+            (round_index, f"Round {current_winners_round_number} {round_type}")
+        )
+
+    return round_indexes_and_names
+
+
+# TODO should this be private-only, and thus should I drop the is_private flag?
+# Could use help/opinions
+def create_round_objects(tournament: Tournament, *, is_private: bool):
```
n8kim1, line 219: similar question here too. note that this method is _only_ called w the private bracket (for now...)

backend/siarnaq/bracket/challonge.py
```diff
@@ -126,21 +126,118 @@ def get_tournament_data(tournament: Tournament, *, is_private: bool):
...
+    round_indexes = _get_round_indexes(tournament, is_private=is_private)
+
+    round_indexes_and_names = []
+
+    if tournament.style == TournamentStyle.SINGLE_ELIMINATION:
+        for round_index in round_indexes:
+            round_indexes_and_names.append((round_index, f"Round {round_index}"))
+        return round_indexes_and_names
+
+    # Otherwise, we are in double elimination
```
j-mao, line 188: Please assert it / throw an Exception if not. This future-proofs if one day, for some reason, we start running Swiss tournaments.

backend/siarnaq/bracket/challonge.py
```diff
@@ -126,21 +126,118 @@ def get_tournament_data(tournament: Tournament, *, is_private: bool):
...
+    current_winners_round_number = None
+    for position in range(len(round_indexes)):
+        round_index = round_indexes[position]
```
j-mao, line 191:
```suggestion
    for position, round_index in enumerate(round_indexes):
```

backend/siarnaq/bracket/challonge.py
j-mao: I think for this function, doesn't make sense to keep `is_private` as a kwarg, because creating round objects is an operation which intuitively operates on the tournament, not private/public brackets.

backend/siarnaq/bracket/challonge.py
j-mao: This function should probably keep `is_private` around, since it queries a specific bracket. Also fix typo:
```suggestion
def _get_round_indexes_and_names(tournament: Tournament, *, is_private: bool):
```

backend/siarnaq/bracket/challonge.py
j-mao: Same as below, probably keep this kwarg, since it changes which bracket gets queried (even if currently useless).

j-mao (approved):

PR #629, opened 2023-01-31
Support mass resume download

### LINKED ISSUE #64: Create resume downloader, which should have only ASCII in our produced filenames
"we really need to enforce ascii… **special chars in the filenames are making the zip not want to extract** (at least on mac)" - Jerry

Yes, I know frontend-only does not prevent people from breaking our system via odd API calls. But it's still a good deterrent, until there's backend code to do this
Also, **we should note this requirement to users, too, in the instructions**

j-mao:
Closes #64.

### DIFF
--- a/backend/siarnaq/api/episodes/admin.py
+++ b/backend/siarnaq/api/episodes/admin.py
@@ -1,5 +1,6 @@
 import structlog
 from django.contrib import admin, messages
+from django.http import HttpResponseRedirect
 from django.utils.html import format_html
 
 from siarnaq.api.compete.models import Match
@@ -10,8 +11,14 @@ from siarnaq.api.episodes.models import (
     Tournament,
     TournamentRound,
 )
+from siarnaq.api.user.models import User
 
 logger = structlog.get_logger(__name__)
+
+
+@admin.action(description="Export all submitted resumes")
+def export_resumes(modeladmin, request, queryset):
+    return HttpResponseRedirect(User.objects.export_resumes(episodes=queryset))
 
 
 class MapInline(admin.TabularInline):
@@ -23,6 +30,7 @@ class MapInline(admin.TabularInline):
 
 @admin.register(Episode)
 class EpisodeAdmin(admin.ModelAdmin):
+    actions = [export_resumes]
     fieldsets = (
         (
             "General",
--- a/backend/siarnaq/gcloud/titan.py
+++ b/backend/siarnaq/gcloud/titan.py
@@ -19,7 +19,9 @@ def request_scan(blob: storage.Blob) -> None:
     blob.patch()
 
 
-def get_object(bucket: str, name: str, check_safety: bool) -> dict[str, str | bool]:
+def get_object(
+    bucket: str, name: str, check_safety: bool, get_raw: bool = False
+) -> dict[str, str | bytes | bool]:
     """
     Retrieve a file from storage, performing safety checks if required.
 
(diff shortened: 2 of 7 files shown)


backend/siarnaq/gcloud/titan.py
```diff
@@ -49,6 +58,11 @@ def get_object(bucket: str, name: str, check_safety: bool) -> dict[str, str | bo
     blob = client.bucket(bucket).get_blob(name)
     match (check_safety, blob.metadata):
         case (False, _) | (True, {"Titan-Status": "Verified"}):
+            if get_raw:
+                return {
```
n8kim1, line 62: for consistency, put a logging thing here?

backend/siarnaq/api/episodes/admin.py
```diff
@@ -10,10 +16,49 @@
...
+        with ZipFile(f, "w") as archive:
+            rank, last_rating = 0, None
+            for user in users:
+                if user.rating != last_rating:
+                    rank, last_rating = rank + 1, user.rating
+                rank_str = "rank-" + str(rank).zfill(rank_len)
+                user_str = user.first_name + "-" + user.last_name
+                if not user_str.isascii():
+                    user_str = "nonascii-user"
+                fname = f"{rank_str}-{user_str}.pdf"
```
n8kim1, line 47: move fname creation into the if block. Also, consider moving the "check rating increment" into the if block too?

n8kim1 (commented):
Code looks good! pointed out some nits but no need. 
More importantly tho I still need to thoroughly test

backend/siarnaq/api/episodes/admin.py
```diff
@@ -10,10 +16,49 @@
...
+                resume = titan.get_object(
+                    bucket=settings.GCLOUD_BUCKET_SECURE,
+                    name=user.profile.get_resume_path(),
+                    check_safety=False,  # TODO: actually check safety, see #628
+                    get_raw=True,
+                )
+                if resume["ready"]:
```
n8kim1, line 54: throw something otherwise, to indicate failure?
feel free to do this later too

n8kim1:
tested on staging, worked smoothly. I don't really trust that I caught everything tho..let's try prod sometime?

n8kim1 (changes requested):
file too large in practice / prod. pls find workaround. (eg, store in gdrive and produce a link????)

backend/siarnaq/gcloud/titan.py
n8kim1: bump @j-mao

backend/siarnaq/api/episodes/admin.py
n8kim1: bump (well in the new place but yeah)

deploy/galaxy/main.tf
```diff
@@ -42,6 +42,23 @@ resource "google_storage_bucket" "secure" {
...
+resource "google_storage_bucket" "ephemeral" {
+  name = "mitbattlecode-${var.name}-ephemeral"
+
+  location      = var.gcp_region
+  storage_class = "STANDARD"
+  labels        = merge(var.labels, {component="storage"})
+
+  lifecycle_rule {
+    condition {
+      age = 1
```
n8kim1, line 54: its nice that the zip dies after a day but...hm...
It's a direct link to download right? Users wouldn't be saving the _link_ for any real reason?
If so then yay
(if not then...if they try to go back and get hit w a cryptic error or smth, might be annoying; would like to try to find a way to explain/handle better)

n8kim1 (approved):
see comments, altho if you're too burnt out, just go ahead and merge

backend/siarnaq/api/episodes/admin.py
n8kim1: @j-mao

j-mao: unsure how to do cleanly rn- raising exceptions just cause http 500 + a crash email instead of a proper exit.

would rather save for later because not high priority

deploy/galaxy/main.tf
j-mao: they shouldn't save the link- in fact, they don't actually need to see the link ever
browsers just download the file and that's the end of it

j-mao: if they lose the file, too bad, go and download it again

backend/siarnaq/gcloud/titan.py
j-mao: this log seems unimportant

n8kim1 (approved):

PR #655, opened 2023-06-27
Add frontend lint / tests to ci workflow

acrantel:
makes frontend2 tests / linting run on each pr.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,13 +43,16 @@ jobs:
             ${{ runner.os }}-build-
             ${{ runner.os }}-
       - name: Install package dependencies
-        run: cd frontend2 && npm install
+        run: npm install
+        working-directory: ./frontend2
       - name: Check linting and formatting
         # Custom script for checking the linting and formatting being in place
-        run: cd frontend2 && npm run lint
+        run: npm run lint
+        working-directory: ./frontend2
       # Run test cases and this could ensure minimum coverage as well if set
       - name: Execute test cases
-        run: cd frontend2 && npm run test
+        run: npm run test
+        working-directory: ./frontend2
   lint:
     name: Linter (pre-commit)
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,38 @@ env:
   CACHE_NUMBER: 1
 
 jobs:
+  unit-test-frontend:
+    # Defines the type of runner the job runs on
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout to the repository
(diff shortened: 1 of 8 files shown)


lowtorola (approved):

n8kim1 (dismissed):
so nice to have

n8kim1:
do we ever plan on renaming frontend2 to frontend and then frontend to "old" or smth? if so, we'll have to make careful note to change this folder too, cuz it's not really a directory name in the same way in code as everything else

to be fair, i think that errors will expose this too

.github/workflows/ci.yml
```diff
@@ -17,6 +17,39 @@ env:
...
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          restore-keys: |
+            ${{ runner.os }}-build-${{ env.cache-name }}-
+            ${{ runner.os }}-build-
+            ${{ runner.os }}-
+      - name: Install package dependencies
+        run: cd frontend2 && npm install
+      - name: Check linting and formatting
+        # Custom script for checking the linting and formatting being in place
+        run: cd frontend2 && npm run lint
```
j-mao, line 49: See the `working-directory` key in the example [here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsrun).

j-mao:
GitHub workflow looks good to me! [modulo nit about using `working-directory` instead of `cd blah &&`] Thanks so much for doing this :tada:



> do we ever plan on renaming frontend2 to frontend and then frontend to "old" or smth? if so, we'll have to make careful note to change this folder too, cuz it's not really a directory name in the same way in code as everything else

This can be done already, can't it? If I understand correctly this branch has zero dependency on the contents of `frontend` so the old stuff can be safely dropped and this directory renamed on the branch. Whenever you need to peek at old stuff, you can still find it in version control on `main` until this big feature branch gets merged.

acrantel:
> GitHub workflow looks good to me! [modulo nit about using `working-directory` instead of `cd blah &&`] Thanks so much for doing this 🎉
> 
> > do we ever plan on renaming frontend2 to frontend and then frontend to "old" or smth? if so, we'll have to make careful note to change this folder too, cuz it's not really a directory name in the same way in code as everything else
> 
> This can be done already, can't it? If I understand correctly this branch has zero dependency on the contents of `frontend` so the old stuff can be safely dropped and this directory renamed on the branch. Whenever you need to peek at old stuff, you can still find it in version control on `main` until this big feature branch gets merged.

will put this into our roadmap

acrantel:
added a recommended plugin to eslint and required semicolons.

n8kim1 (approved):

PR #658, opened 2023-07-06
Table Component/Rankings Page

lowtorola:
- Adds fully generic `BattlecodeTable.tsx` element, with support for pagination via an optional `BattlecodeTableBottomElement.tsx` bottom element!
- Creates Rankings page for viewing a list of all our lovely teams!


lowtorola:
TODOS:
- [ ] Make table actually 80% of window width
- [ ] Only light mode on table 😢 
- [ ] Add loading safety to `onPage` for `BattlecodeTableBottomElement`
- [ ] Add search

lowtorola:
Consider alternative to useEffect for query... maybe function called in useEffect first page render? then called on page/search?

lowtorola:
BOOM 💥

acrantel (commented):
great job! left comments about async/await preferences and moving the spinner into components/elements

should be good to merge after we address those, so just rerequest me for review after you make changes

acrantel:
image
also, the BattlecodeTableBottomElement shows every page from 1-[num_pages]. let's make it so that it cuts off the overflow at some point

one way we could display page nav at the bottom of the table is this (we would probably display more page numbers around the page in our actual implementation):
if our page # is 1
* `1, 2, 3, 4, 5, 6, 7, 8, 9, 10, ... 200`
if our page  # is 12
* `1 ... 9, 10, 11, 12, 13, 14, 15 ... 200`
if our page # is 200
* `1 ... 196, 197, 198, 199, 200`

frontend2/src/components/BattlecodeTableBottomElement.tsx
```diff
@@ -20,6 +20,38 @@ const BattlecodeTableBottomElement: React.FC<TableBottomProps> = ({
   const backDisabled = currentPage <= 1;
   const forwardDisabled = currentPage >= pageCount;
 
+  function getPageNumbers(): Array<string | number> {
```
acrantel, line 23: could you add a comment here describing what this function does and its the return value? I see it returns strings and numbers but i'm not sure what that means

frontend2/src/components/BattlecodeTableBottomElement.tsx
```diff
@@ -20,6 +20,38 @@ const BattlecodeTableBottomElement: React.FC<TableBottomProps> = ({
...
+          .concat(
+            Array.from(
+              { length: MAX_PAGES - 4 },
+              (_, idx) => idx + currentPage - 5
+            )
+          )
+          .concat(["...", pageCount]);
+      }
+    } else if (pageCount < 1) {
+      return ["1"];
```
acrantel, line 49: looks like we're returning a string containing 1 here instead of the number 1. is this intentional?

acrantel (approved):
tested the new table bottom element and it looks great! left some comments regarding the return type / documentation of the `getPageNumbers` function, but once that is addressed then this LGTM

PR #662, opened 2023-07-29
Add DocumentationPage component (for Resources / Quickstart page) and make header/sidebar fixed on scroll

apollo1291:
Adds the `DocumentationPage` component, which takes markdown text and renders it with react-markdown. 
* Used for Resources and Quickstart page

Also, makes the header and sidebar fixed

**Note: This PR does not contain the correct text for the quickstart / resources pages. The text will be added later b/c it's lower priority**

image


frontend2/src/views/Rankings.tsx
```diff
@@ -64,7 +64,7 @@ const Rankings: React.FC = () => {
   }, [searchQuery, page]);
 
   return (
-    <div className="mb-20 ml-10 flex w-full flex-col">
+    <div className="flex h-full w-full flex-col overflow-auto p-6">
```
acrantel, line 67: we have to use padding instead of margins here, otherwise it will overflow the page width because of the combined `w-full` and `ml-10`

frontend2/package.json
```diff
@@ -17,6 +17,7 @@
     "react-dom": "^18.2.0",
     "react-hook-form": "^7.45.1",
     "react-router-dom": "^6.13.0",
+    "react-markdown": "^8.0.7",
```
acrantel, line 20: in the future, we may consider switching to `marked` or `remarkable`, because they are about 1/3 the size of react-markdown.
we'll just need to figure out how to replace `<a>`s that link to urls within play.batlecode.org with `<Link>`

frontend2/src/views/QuickStart.tsx
```diff
@@ -1,7 +1,9 @@
 import React from "react";
+import DocumentationPage from "../components/DocumentationPage";
+import { BC23_QUICKSTART } from "../content/bc23";
 
 const QuickStart: React.FC = () => {
```
lowtorola, line 5: Is this view going to always bring up `bc23`'s Quickstart? Or are we planning to use EpisodeContext on this page (e.g. `bc22`,`bc23`) that indexes and retrieves `<episodeId>_QUICKSTART` for the `Quickstart` component?

frontend2/src/views/Register.tsx
```diff
@@ -75,7 +75,7 @@ const Register: React.FC = () => {
         {
           // TODO: replace this with our custom notification component
```
lowtorola, line 76: Is this on our roadmap?

frontend2/src/views/Resources.tsx
```diff
@@ -0,0 +1,9 @@
+import React from "react";
+import { BC23_RESOURCES } from "../content/bc23";
+import DocumentationPage from "../components/DocumentationPage";
+
+const Resources = (): JSX.Element => {
```
lowtorola, line 5: Ditto the comment above about this page eventually rendering the appropriate year's resources based on the URL's `episodeId`? If that was already the plan then disregard.

lowtorola (approved):
LGTM! Left a few comments about future plans for these markdown pages but looks really good! 🎉

frontend2/src/views/QuickStart.tsx
acrantel: yup, this will be a later change where we'll get the episode id and display that episode's quickstart

frontend2/src/views/Register.tsx
acrantel: Yup!

frontend2/src/views/Resources.tsx
acrantel: yeah, we'll render the proper year later. that change will likely come in the same pr as the change that adds the actual 2023 / 2024 episode specific text

PR #664, opened 2023-08-11
Api Cleanup

lowtorola:
- Changed API/Auth to be exported functions in separate files.
- Accidentally changed ~5000 lines by running format (oopsie).

:rocket:

### DIFF
--- a/frontend2/.eslintrc.js
+++ b/frontend2/.eslintrc.js
@@ -18,7 +18,7 @@ module.exports = {
     "*.json",
     ".eslintrc.js",
     "tailwind.config.js",
-    "**/types/**",
+    "src/utils/types",
   ],
   parserOptions: {
     project: "tsconfig.json",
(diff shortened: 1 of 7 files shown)


lowtorola:
halp @acrantel my lint seems to be borked 😢

acrantel (changes requested):
thanks for doing this!

what command did you run that formatted all the autogenerated files? if it was `npm run format` or `lint` then we'll need to update the ignores in `eslintrc.js` and `.prettierignore` . it should be fairly straightforward, just put the path to the autogenerated folder in the ignore lists

we shouldn't commit the format changes to the autogenerated files, here is a stackoverflow solution for how to remove them from your pr: https://stackoverflow.com/a/39460058

frontend2/.gitignore
```diff
@@ -11,6 +11,9 @@
 # production
 /build
 
+# auto-generated files
+/src/utils/types
```
acrantel, line 15: we shouldn't add this change, we want these types to stay in our git repo

lowtorola: Oh whoops that's right this should be in `.prettierignore` 🤦

acrantel (approved):
lgtm! 🎉 :shipit:

PR #667, opened 2023-08-19
Change api generation

acrantel:
API generation changes
* switch from typescript-jquery to typescript-fetch so that our generated enums can be strings.
* use snake_case in generated types instead of camelCase (to match what backend expects)
* various changes in backend files to make the generated schema more accurate

Frontend API client changes
* refactored api.ts and auth.ts into multiple files in utils/api. 
* deleted some legacy functions (especially in compete.ts) 
* made all our api functions use the generated library instead of jquery

Miscellaneous:
* changed the bat file for generating types: `generate-types.bat`
* updated our precommit hooks to exclude the types folder from formatting (see `.pre-commit-config.yaml`)
* deleted jquery from package.json

**review requests**
* You can ignore the types folder when reviewing these files.
* please check that the backend schema changes make sense / are accurate


n8kim1 (approved):
comments are all things to doublecheck, etc -- no feedback that would require a second validation from me 
looks good! happy if this just gets revisions and then is merged

(unless we have PRs dismiss on stale)

frontend2/src/utils/api.ts
```diff
@@ -297,31 +300,33 @@ export const getAllUserTournamentSubmissions = async (
...
-      ...user,
-      profile: {
-        ...defaultUser.profile,
-        ...user.profile,
-        gender: user.profile.gender as unknown as models.GenderEnum,
-        country: user.profile.country as unknown as models.CountryEnum,
-      },
-    })
-  ).body;
+  return await API.apiUserUCreate({ userCreate });
```
lowtorola, line 329: Just want to confirm that this new generated api function call (await foo() vs. await foo().body) returns the same model as the old one so that no breaking bugs pop up in the already-built UI 😁

frontend2/src/utils/types/.openapi-generator/FILES
```diff
@@ -1,65 +1,64 @@
...
-model/UserPrivate.ts
-model/UserProfilePrivate.ts
-model/UserProfilePublic.ts
-model/UserPublic.ts
-model/UserResume.ts
-model/models.ts
-variables.ts
+models/Autoscrim.ts
+models/ClassRequirement.ts
+models/CountryEnum.ts
```
lowtorola, line 7: Just to confirm (in similar vein to other comment) that this folder structure change will persist forever after generate types is run so that no problems pop up

lowtorola (changes requested):
Looks really awesome! Left a few comments about stability with future typegen runs but if those are non-issues then ship it!

frontend2/src/utils/types/.openapi-generator/FILES
```diff
@@ -1,65 +1,64 @@
...
-model/UserCreate.ts
-model/UserPassed.ts
-model/UserPrivate.ts
-model/UserProfilePrivate.ts
-model/UserProfilePublic.ts
-model/UserPublic.ts
-model/UserResume.ts
-model/models.ts
-variables.ts
+models/Autoscrim.ts
```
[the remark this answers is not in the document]
acrantel: this was an automatic change

frontend2/src/utils/types/.openapi-generator/VERSION
```diff
@@ -1 +1 @@
-6.6.0
+6.6.0
```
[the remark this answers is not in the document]
acrantel: yup, everything in /types are autogenerated

frontend2/src/utils/api.ts
```diff
@@ -1,8 +1,9 @@
-import { ApiApi } from "./types/api/ApiApi";
+import { ApiApi } from "./types/apis/ApiApi";
```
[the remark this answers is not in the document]
acrantel: this was an automatic change by the generator - if we continue using typescript-fetch, it shouldn't change again

frontend2/src/utils/api.ts
acrantel: yup, tht's correct

frontend2/src/utils/types/.openapi-generator/FILES
acrantel: yup, I made no changes to the /types folder after generating

acrantel:
this PR is blocked by https://github.com/battlecode/galaxy/pull/671.

acrantel (commented):
still todo: update all references to the API in our frontend components

backend/siarnaq/api/user/views.py
```diff
@@ -75,6 +75,14 @@ def me(self, request):
                 serializer.save()
                 return Response(serializer.data)
 
+    @extend_schema(
+        responses={
+            status.HTTP_200_OK: {
+                "type": "object",
+                "additionalProperties": {"$ref": "#/components/schemas/TeamPublic"},
+            }
```
acrantel, line 83: this schema dict describes the response as an object with arbitrary keys with each value conforming to the schema TeamPublic (which is generated from TeamPublicSerializer)

n8kim1:
seems good, would wait for lowell though

backend/siarnaq/settings.py
```diff
@@ -510,3 +510,18 @@ def dropper(logger, method_name, event_dict):
         raise structlog.DropEvent
 
     structlog.configure(processors=[dropper])
+
+SPECTACULAR_SETTINGS = {
+    # Split components into request and response parts where appropriate
+    # This setting is highly recommended to achieve the most accurate API
+    # description, however it comes at the cost of having more components.
+    "COMPONENT_SPLIT_REQUEST": True,
```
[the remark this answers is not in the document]
lowtorola: 🎉🎉

lowtorola (approved):
Looks fantastic! Love the split API! Next step,  implementing team ranking & ranking history!

PR #670, opened 2023-08-20
Upgrade prettier, add some ui elements, add the Register page.

acrantel:
Adds the register page.
UI elements added / styled (these have all been tested, see the Register.tsx page):
* Button
* FormError
* FormLabel
* Icon
* Input
* SelectMenu

Prettier changes: 
* Adds [@headlessui/tailwindcss](https://www.npmjs.com/package/@headlessui/tailwindcss) to style headless ui components based on state.
* Adds prettier-plugin-tailwindcss to sort tailwind classes
* Upgrades prettier so that prettier-plugin-tailwindcss works
* Removes prettier precommit hook because it conflicts with having `prettier-plugin-tailwindcss` listed in `prettierrc.json`
* Formats the entire project with the new prettier setup (it added a lot of commas lmao)

**Before testing, make sure to run `npm install` to get the updated packages.**

image


.pre-commit-config.yaml
```diff
@@ -8,13 +8,6 @@ repos:
         args: ["--fix=lf"]
       - id: trailing-whitespace
 
-  - repo: https://github.com/pre-commit/mirrors-prettier
-    rev: v2.7.1
-    hooks:
-      - id: prettier
-        types_or: [css, html, javascript, json, jsx, markdown]
-        exclude: ^frontend/public/assets
-
```
acrantel, line 17: I'm removing this precommit hook because it does the same thing as `npm run format`, and it conflicts with having the `prettier-plugin-tailwindcss` in `.prettierrc.json`.

To use plugins with precommit prettier, you're supposed to add this: 
```
additional_dependencies:
  - prettier
  - prettier-plugin-tailwindcss
```
to the precommit config, but this still errors if `"plugins": ["prettier-plugin-tailwindcss"]` is included in `prettierrc.json`.

I would prefer to have `"plugins": ["prettier-plugin-tailwindcss"]` in `prettierrc.json` than have this pre-commit hook.

n8kim1: we still have a check in github actions for PRs right? that cathes the same things
just double-checking (cuz I'm in a rush oop i can check later)

frontend2/src/components/elements/Icon.tsx
```diff
@@ -0,0 +1,85 @@
...
+  clock: ClockIcon24,
+  user_group: UserGroupIcon24,
+  arrow_up_tray: ArrowUpTrayIcon24,
+  play_circle: PlayCircleIcon24,
+  chevron_down: ChevronDownIcon24,
+  check: CheckIcon24,
+  information_circle: InformationCircleIcon24,
+};
+
+const icons20 = {
```
n8kim1, line 47: we need both 24 and 20? would it be worth it to cut down and only need to import one of them?

frontend2/src/components/elements/SelectMenu.tsx
```diff
@@ -0,0 +1,90 @@
+import React, { Fragment, useMemo, useState } from "react";
```
n8kim1, line 1: "Check warning on line 1 in frontend2/src/components/elements/SelectMenu.tsx

GitHub Actions
/ Frontend linting / unit tests

'useState' is defined but never used"

GHA said that ^ woahhh
but yeah, drop it for good snytax?

frontend2/src/components/elements/Icon.tsx
j-mao: (jumping in here as someone who hasn't been paying attention and who has just read some email digests)

I think it's ok to have both, because it makes sense to be able to tailor/change icon-sets for different episodes of Battlecode. But, this is conditioned on:

- The setup being something easily extensible to *n* episodes of Battlecode for large *n*.
- I think it's ok if this extensibility means defining a new dict/mapping for each episode, because that's unavoidable when each episode could be different.
- However, it should also be implementable without changing the downstream rendering code, so [this](https://github.com/battlecode/galaxy/pull/670/files/444aa6f45c9ef9425827eb8ae2217fc4690f8504#diff-b9386fd95d3bd66ff698696964d7a7e70d9acdf0199982091a61340678d36736R62) looks a bit more dubious. Perhaps it could be a single mapping that maps each episode name to one of these dicts?

Separately, curious how easy it would be to use this framework to add new menu items in the future with new icons? (possibly the answer is "very easy", I didn't read the code at all)

lowtorola (commented):
LGTM pending resolution of Jerry & Nathan's comments 🙌

frontend2/src/components/elements/Icon.tsx
n8kim1: (oops, 24/20 px not episode?)

frontend2/src/views/Register.tsx
```diff
@@ -1,7 +1,172 @@
...
+  const onSubmit: SubmitHandler<CreateUserInput> = async (data) => {
+    if (gender === undefined || country === undefined) {
+      return;
+    }
+    try {
+      const newUser = await Auth.register(data);
+      login(newUser);
+      console.log("logged in successfully");
+    } catch (error) {
+      console.log("failure to register", error);
```
n8kim1, line 40: Consider showing a user-facing error, even just propagating the backend message up to the frontend is fine. (errors where nothing seems to happen would be annoying to user)

n8kim1: and it's not clear to me that the error does show, especially if it's in the login method? If it does feel free to ignore me, I haven't locally tested yet

frontend2/src/components/elements/Icon.tsx
n8kim1: ah, if either size gets used for different purposes, then sounds good
I'm a little wary of the extra overhead here -- if we want to add a new icon, then we'd have to import both sizes, and make sure the correct size gets used. But perhaps this is easier than I think. Most of all though, we could always come back to this later

frontend2/src/views/Register.tsx
acrantel: I'll do this in a separate PR, as this PR is already pretty big. In that PR I'm also planning on creating a reusable component for displaying these errors.

n8kim1 (approved):
overall looks good, do see comments though!

frontend2/src/views/Register.tsx
n8kim1: oooo a reusable error comment, niceeee

acrantel: Actually for this PR I can make a super simple error message and replace it in the later PR

frontend2/src/components/elements/Icon.tsx
j-mao: ah fair, naming wasn't entirely clear but makes sense now

PR #671, opened 2023-08-23
Fix team update/create bugs

acrantel:
Requesting that Jerry review this if possible because he most recently touched this code.

This PR fixes a bug that allows users to change what episode their team belongs to, and if their team has a report or not.

Testing:
* Tested creating and updating a team locally.


lowtorola (approved):
LGTM! But I agree with potentially awaiting review from Jerry since he worked on this!

j-mao (approved):
looks reasonable to me

PR #672, opened 2023-08-25
Add login page

acrantel:
Adds the login page.

Testing:
* Try logging in with invalid username / pwd. Error message should appear
* Try logging in with correct username / pwd. Login should succeed and should redirect to the home page of the most recently accessed episode.
* Try going to the /login page while already logged in. It should redirect you to the home page of the most recently accessed episode.

Note: testing is blocked on https://github.com/battlecode/galaxy/pull/667.

image

### DIFF
--- a/frontend2/src/views/Login.tsx
+++ b/frontend2/src/views/Login.tsx
@@ -1,7 +1,105 @@
-import React from "react";
+import React, { useContext, useEffect, useState } from "react";
+import Input from "../components/elements/Input";
+import Button from "../components/elements/Button";
+import { login as apiLogin } from "../utils/api/auth";
+import { type SubmitHandler, useForm } from "react-hook-form";
+import { FIELD_REQUIRED_ERROR_MSG } from "../utils/constants";
+import { Link, useNavigate } from "react-router-dom";
+import { EpisodeContext } from "../contexts/EpisodeContext";
+import { useCurrentUser, AuthStateEnum } from "../contexts/CurrentUserContext";
+import { type Maybe } from "../utils/utilTypes";
+import { getUserUserProfile } from "../utils/api/user";
+
+interface LoginFormInput {
+  username: string;
+  password: string;
+}
 
 const Login: React.FC = () => {
-  return <p>login page</p>;
+  const { register, handleSubmit } = useForm<LoginFormInput>();
+  const { episodeId } = useContext(EpisodeContext);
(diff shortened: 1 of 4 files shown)


n8kim1:
test-ready now?

frontend2/src/views/Login.tsx
```diff
@@ -1,7 +1,105 @@
...
+    }
+  }, [authState]);
+
+  const onSubmit: SubmitHandler<LoginFormInput> = async (data) => {
+    try {
+      setLoginError(undefined);
+      await apiLogin(data.username, data.password);
+      login(await getUserUserProfile());
+    } catch (error) {
+      console.log(error);
```
lowtorola, line 39: Do you think it would be useful to send the specific (i.e. username/pass incorrect) errors to the form elements? Thinking if our jwt errors are a consistent format, then we can use that to determine which is incorrect? If not, not needed :+1:

frontend2/src/views/Login.tsx
```diff
@@ -1,7 +1,105 @@
...
+      <form
+        // eslint-disable-next-line @typescript-eslint/no-misused-promises
+        onSubmit={handleSubmit(onSubmit)}
+        className="flex w-11/12 flex-col gap-5 rounded-lg bg-gray-100 p-6 shadow-md sm:w-[350px]"
+      >
+        <h3 className="text-center text-xl font-light text-gray-700">
+          Log in to Battlecode
+        </h3>
+        {
+          // TODO: replace this with our custom notification component
```
lowtorola, line 60: Maybe should make an issue for this so the TODO can track it?

lowtorola (approved):
Left a few comments, but all in all looks really good! Didn't get a chance to test on my local (at work while reviewing) but if you've tested to a level you're comfortable with merge it! 🎉

frontend2/src/views/Login.tsx
acrantel: The errors are of a consistent type: `ResponseError` if the server returns a status <200 or >=300, which contains the `response` and a `msg`. 

unfortunately the server doesn't tell us if the username or password is incorrect, so we can't display that info to the user

frontend2/src/views/Login.tsx
acrantel: I'll put this into our roadmap spreadsheet

PR #673, opened 2023-08-31
Add tooltip

acrantel:
image

example usage:
```
<Tooltip text="this is part of the tooltip.">
  <p className="text-pink-500">the default location for a tooltip is top</p>
</Tooltip>
<Tooltip location="right" text="This is an incredibly long tooltip text that should wrap nicely">
  <Icon name="arrow_up_tray" size="sm" />
</Tooltip>
<Tooltip location="left" text="This is an incredibly long tooltip text that should wrap nicely">
  <Icon name="map" size="sm" />
</Tooltip>
<Tooltip location="bottom" text="short tooltip">
  <Icon name="map" size="sm" />
</Tooltip>
```

### DIFF
--- a/frontend2/src/components/elements/Tooltip.tsx
+++ b/frontend2/src/components/elements/Tooltip.tsx
@@ -1,0 +1,55 @@
+import React from "react";
+
+// Example usage
+// <Tooltip content="This will be displayed inside the tooltip" location="left">
+//   <p>Hovering this makes the tooltip appear</p>
+// </Tooltip>
+
+interface TooltipProps {
+  children?: React.ReactNode;
+  content: string;
+  delay?: number;
+  location?: "top" | "bottom" | "left" | "right";
+}
+
+const TOOLTIP_CLASSES = {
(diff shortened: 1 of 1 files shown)


n8kim1:
wow fancy demos

wait is this homemade? if so, niceeeee, and might suggest adding the  code that generated that screenshot to the repo
(or add that code snippet that you put in PR desc)

if not homemade would suggest throwing in a link as comment too

lowtorola (approved):
Looks amazing! My one thought would be maybe we could rename `content` to `title`? Since it is limited to being a text input. LGTM though, I'll put this in use asap! :tada:

acrantel:
> My one thought would be maybe we could rename content to title?

Hmm, I think the text in the tooltip isn't really a title. What if we rename `content` to `text`?

lowtorola:
> > My one thought would be maybe we could rename content to title?
> 
> Hmm, I think the text in the tooltip isn't really a title. What if we rename `content` to `text`?

Works for me!

PR #677, opened 2023-09-12
Scrimmaging Page

lowtorola:
Adds a scrimmaging page!


frontend2/src/views/Scrimmaging.tsx
```diff
@@ -0,0 +1,340 @@
...
+    setTourneyLoading(true);
+    setTourneyData((prev) => ({ count: prev?.count ?? 0 }));
+    try {
+      const result = await getTournamentMatches(
+        episodeId,
+        undefined,
+        undefined,
+        undefined,
+        tourneyPage,
+      );
```
lowtorola, line 104: Todo: need to send this to the state!

frontend2/src/views/Scrimmaging.tsx
```diff
@@ -0,0 +1,340 @@
...
+      <div className="justify-right mb-1 flex flex-row space-x-4">
+        <h1 className="mb-2 text-2xl font-bold leading-7 text-gray-900">
+          Scrimmage History
+        </h1>
+        <Button
+          disabled={scrimsLoading}
+          label="Refresh!"
+          variant="dark"
+          onClick={() => {
+            setQueryParams({ ...queryParams, scrimsPage: "1" });
```
lowtorola, line 245: switch these so the if statement goes first and then returns!

frontend2/src/views/Scrimmaging.tsx
```diff
@@ -0,0 +1,340 @@
...
+
+      <div className="mb-4 w-5/6">
+        <BattlecodeTable
+          data={tourneyData?.results ?? []}
+          loading={tourneyLoading}
+          columns={[]}
+          bottomElement={
+            <BattlecodeTableBottomElement
+              totalCount={tourneyData?.count ?? 0}
+              pageSize={10}
```
lowtorola, line 324: should make this an optional prop in `BattlecodeTableBottomElement.tsx` and default to 10. Maybe even don't allow this to be changed and add page sizing to be changed in db calls in a future PR?

lowtorola:
TODO: request scrimmages against a team and implement "incoming scrimmage requests" section

lowtorola:
Convo at meeting 9/30:
- Want to refactor `BattlecodeTable` usages into e.g. `RecentTournamentMatchesTable.tsx`
- ^^ will require new subfolders e.g. `Scrimaging` in `components`
- Want to refactor new `InboxOutbox` component
- Move fetch functions into useEffects, set loading above functions, and use destructors to prevent race conditions (see EpisodeProvider)

frontend2/src/views/Scrimmaging.tsx
lowtorola: TODO in future PR

lowtorola:
In order to be truly done, this page will need the following dialogs/buttons:
- [ ] ACCEPT scrimmage
- [ ] REJECT scrimmage
- [ ] CANCEL scrimmage
- [ ] match replay button

frontend2/src/views/Scrimmaging.tsx
```diff
@@ -0,0 +1,542 @@
...
+  };
+
+  interface QueryParams {
+    inboxPage: number;
+    outboxPage: number;
+    teamsPage: number;
+    scrimsPage: number;
+    tourneyPage: number;
+    search: string;
+  }
```
acrantel, line 52: nit: move this interface definition outside the component function b/c it doesn't depend on component state

frontend2/src/views/Scrimmaging.tsx
```diff
@@ -0,0 +1,542 @@
...
+        setInboxLoading(false);
+      }
+    };
+
+    void fetchInbox();
+
+    return () => {
+      isActiveLookup = false;
+    };
+  }, [queryParams.inboxPage]);
```
acrantel, line 129: All of these useEffects also depend on episodeId, so we should add that into the dependencies

frontend2/src/views/Scrimmaging.tsx
```diff
@@ -0,0 +1,542 @@
...
+    search: string;
+  }
+
+  const queryParams: QueryParams = useMemo(() => {
+    return {
+      inboxPage: parseInt(searchParams.get("inboxPage") ?? "1"),
+      outboxPage: parseInt(searchParams.get("outboxPage") ?? "1"),
+      teamsPage: parseInt(searchParams.get("teamsPage") ?? "1"),
+      scrimsPage: parseInt(searchParams.get("scrimsPage") ?? "1"),
+      tourneyPage: parseInt(searchParams.get("tourneyPage") ?? "1"),
```
acrantel, line 60: nit: we can reduce repetition here by using a small function 
`const parsePageParam = (paramName: string) => parseInt(searchParams.get(paramName) ?? "1");`

acrantel (changes requested):
Great work! Left some comments mostly regarding code readability / organization

I think we briefly discussed this on Tuesday, but I just realized that using `setSearchParams` in `Scrimmaging.tsx` will cause the entire page to rerender b/c searchParams changes. 
This issue has more discussion about this rerendering issue: https://github.com/remix-run/react-router/issues/8908

I believe you mentioned that we can directly update window.history instead of using setSearchParams, and I think this would avoid the rerendering issue.

frontend2/src/components/compete/MatchScore.tsx
```diff
@@ -1,27 +1,80 @@
...
-import { type Match, StatusBccEnum } from "../../utils/types/models";
+import {
+  type Match,
+  StatusBccEnum,
+  type MatchParticipant,
+} from "../../utils/types/models";
 
 interface MatchScoreProps {
   match: Match;
+  userTeamId?: number;
```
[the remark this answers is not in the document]
acrantel: ignore this comment, userTeamId may not be the current user's team

acrantel (approved):
This is awesome! Just left two minor requests for change, but once those are addressed feel free to merge

PR #679, opened 2023-09-30
Add team context, update user/episode context

acrantel:
Adds a context to store the user's current team's information.

Updates the user context to reduce rerendering

Updates Episode context to contain full episode information instead of just the id, and adds EpisodeProvider, which encapsulates the logic for loading the full episode information.

- Note that there are two hooks to get episode data: `useEpisode` and `useEpisodeId`. `useEpisode` contains full episode data, and `useEpisodeId` contains the episodeId and a fucntion to set the episode id. I set it up this way to avoid rerenders.
  - components that only require episodeId will not rerender when the full episode data is loaded. ( I documented this in the code as well)

### DIFF
--- a/frontend2/src/App.tsx
+++ b/frontend2/src/App.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React from "react";
 import EpisodeLayout from "./components/EpisodeLayout";
 import Home from "./views/Home";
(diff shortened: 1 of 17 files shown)


lowtorola (approved):
LGTM! Love the useMemo's looks well implemented to me 😄

PR #682, opened 2023-10-03
Move api gen to frontend2, add back old frontend prettier, bump CI node version

acrantel:
- [x] Bump frontend2 CI node version to 18
- [x] Move toplevel api generation files to frontend2"
- [x] ensure linting still runs on old frontend
- [x] regenerate the API because the backend in main was different from the backend in frontend2

This PR also documents the api generation process and adds a list of important scripts to the  `frontend2/README.md `


package-lock.json
acrantel, line 1: Moved the api generation to /frontend2, so these are obselete.

frontend2/schema.yml
```diff
@@ -2486,9 +2486,6 @@ components:
       properties:
         profile:
           $ref: '#/components/schemas/TeamProfilePrivateRequest'
-        episode:
-          type: string
-          minLength: 1
     PatchedUserPrivateRequest:
```
acrantel, line 2489: This change came from rebasing onto main. There was a bugfix that disabled the ability to change an team's episode.

.github/workflows/ci.yml
```diff
@@ -17,8 +17,8 @@ env:
   CACHE_NUMBER: 1
 
 jobs:
-  unit-test-frontend:
-    name: Frontend linting / unit tests
+  unit-test-frontend2:
```
j-mao, line 20: Are you planning to rename this to frontend eventually when the entire branch gets merged?

(at some point it becomes safe to delete the frontend directory entirely?)

.github/workflows/ci.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       - name: Set up NodeJS environment
         uses: actions/setup-node@v3
         with:
-          node-version: '14'
+          node-version: '18.11.0'
```
acrantel, line 30: This is the node version in `environment-dev.yml`

.pre-commit-config.yaml
```diff
@@ -11,6 +11,14 @@ repos:
       - id: trailing-whitespace
         exclude: ^frontend2/src/utils/types
 
+  # This is only used for the old frontend. The prettier check for frontend2 exists in ci.yml
+  - repo: https://github.com/pre-commit/mirrors-prettier
+    rev: v2.7.1
+    hooks:
+      - id: prettier
+        types_or: [css, html, javascript, json, jsx, markdown]
+        exclude: ^(frontend/public/assets|frontend2)
```
j-mao, line 20: Excluding frontend2 -- why do we need this then?

Also- pre-commit is useful for catching things locally before they get rejected by CI, so may recommend keeping if it's DRY.

environment-dev.yml
```diff
@@ -28,6 +28,7 @@ dependencies:
   - python=3.10
   - pyupgrade=3.1.0
   - terraform=1.3.4
+  - openjdk=11
```
j-mao, line 31: Out of curiosity, why 11? This is neither our outdated Java 8 nor the latest.

frontend2/README.md
```diff
@@ -2,6 +2,42 @@
...
 See [docs/onboard.md](docs/onboard.md) for an introduction to the frontend codebase, in order to make changes. This is helpful not only for frontend devs themselves but also for **just a working knowledge of frontend features, even if you're not a frontend dev!**
+
+## Generate API types
+
+We use a bash script to generate API types / functions from our backend code automatically. To run the bash script, ensure you have followed all the setup setps in the top-level directory and in [docs/local-setup.md](docs/local-setup.md).
+
+The following steps assume that you begin in the `galaxy` directory.
+
+1. `cd ./frontend2`
+2. `chmod +x generate_types.sh` (makes the script executable, only needs to be run one time)
```
j-mao, line 42: You can `chmod +x ./generate_types.sh` and commit that change instead of requiring this to be done by every dev. The x bit on each file is tracked by git.

j-mao (commented):
Dropping some thoughts but feel free to disregard

.github/workflows/ci.yml
acrantel: We're planning on merging frontend2 branch into main before onboarding people, so we'll keep things in the frontend2 folder for now. We can rename the tests and everything else when frontend2 is ready to be the frontend

.pre-commit-config.yaml
acrantel: this is in prep for merging into main, so that the old frontend still has linting checks

I excluded the frontend2 folder because frontend2 has a prettier tailwindcss extension and I can't get it to work with precommit (which is why frontend2 prettier checks are in ci.yml instead of precommit)

frontend2/README.md
acrantel: oh great! will remove that command from the readme

environment-dev.yml
acrantel: It was what was automatically installed when I added "openjdk" with conda. 

I believe openjdk 8 doesn't work with our engine anyways, so that's why  it's not at java 8.

acrantel: Ah actually it looks like openapigenerator has a minimum of java 11 and may require java 11 for mac users

https://github.com/OpenAPITools/openapi-generator#13---download-jar

j-mao (approved):
seems legit ig

PR #689, opened 2023-10-18
Add team sidebar display conditions

sri240:
If the user is not logged in, the entire TEAM MANAGEMENT section of the sidebar should not appear
If the user is logged in and does not have a team, the Submissions and Scrimmaging sections should not appear
If the user is logged in and has a team, the entire TEAM MANAGEMENT section should appear


acrantel (changes requested):
Overall looks good! Left a comment about nullness / undefinedness.

It turns out that sidebar is the only component right now with tests, so we'll need to update those tests before merging this PR. I believe the change we'll need to make is to add a `<CurrentUserContext.Provider>` and `<CurrentTeamContext.Provider>` to wrap the `<Sidebar>` in the tests. You can find the test at components/sidebar/__test__/Sidebar.test.tsx

acrantel (approved):
looks good to me! feel free to merge :)

PR #690, opened 2023-10-21
Add notes from onboarding meeting

n8kim1:
Begins adding notes for onboarding! 
Happy to add more as needed, but don't think it should be in _this_ PR.

Covers most things that we covered during our first meeting, _except for battlecode-specific frontend files_, cuz new devs have written frontend code already, we never recorded the presenting here, I had to tune out during it, and I don't know it myself oops. Better for someone else to do. (Also I don't want this otherwise-complete PR hanging)

Everything here is still a good reference, for new devs, for other devs that have to wade into webinfra, and for next year.


docs-general/onboard.md
```diff
@@ -0,0 +1,57 @@
...
+
+![System Diagram](system-diagram.png)
+
+## The big stuff
+
+Here, we introduce the systems through tracking the flow of information.
+
+First, a competitor uses their own computer to load up `play.battlecode.org`, thus visiting the _frontend_. Through the competitor's interactions, the frontend makes Internet requests to the _backend_. The backend allows these requests via the _API_, an interface that the backend defines to communicate with the outside world. Finally, the backend stores and retrieves bulk amounts of data in the _database_.
+
+For viewing general information or managing team/user data, the flow of data generally ends here. However, when a user wants to submit new source code for their bot, or submit a request to scrimmage against another team... these are more involved.
```
acrantel, line 34:
```suggestion
For viewing general information or managing team/user data, the flow of data generally ends here. However, when a user wants to submit new source code for their bot, or request a scrimmage against another team... these are more involved.
```

acrantel (approved):
looks great! left small comments on clarifying what the two pubsubs are

n8kim1:
lmao you can tell which part was the last part i wrote, and which part i
gave up on

thanks for calling that out , needed it
and ty for writing it for me too

PR #692, opened 2023-10-21
Nicole/register kerb field

nicolexxuu:
If (and only if) the user enters MIT in the school field, the registration should also display a Kerberos field


acrantel (approved):
great job! tested and it works :)
once the linting is fixed then we can merge this in

PR #696, opened 2023-11-03
Make leave team button work, add modal component

acrantel:
image

## Testing plan
Clicking the leave team button should display the popup in the screenshot above. Clicking the cancel button should close the modal, and clicking the leave team button should leave the team. 

Clicking outside the modal should also close the modal.

### DIFF
--- a/frontend2/src/components/Modal.tsx
+++ b/frontend2/src/components/Modal.tsx
@@ -1,0 +1,60 @@
+import { Transition, Dialog } from "@headlessui/react";
+import React, { Fragment } from "react";
+
+interface ModalProps {
+  isOpen: boolean;
+  closeModal: () => void;
+  title: string;
+  children?: React.ReactNode;
+}
+
+const Modal: React.FC<ModalProps> = ({
+  isOpen,
+  closeModal,
+  children,
+  title,
+}) => {
+  return (
+    <Transition appear show={isOpen} as={Fragment}>
+      <Dialog as="div" className="relative z-10" onClose={closeModal}>
+        <Transition.Child
+          as={Fragment}
+          enter="ease-out duration-300"
+          enterFrom="opacity-0"
+          enterTo="opacity-100"
+          leave="ease-in duration-200"
+          leaveFrom="opacity-100"
+          leaveTo="opacity-0"
+        >
+          <div className="fixed inset-0 bg-black/25" />
+        </Transition.Child>
+
+        <div className="fixed inset-0 overflow-y-auto">
+          <div className="flex min-h-full items-center justify-center p-4 text-center">
+            <Transition.Child
+              as={Fragment}
+              enter="ease-out duration-300"
+              enterFrom="opacity-0 scale-95"
+              enterTo="opacity-100 scale-100"
+              leave="ease-in duration-200"
+              leaveFrom="opacity-100 scale-100"
+              leaveTo="opacity-0 scale-95"
+            >
+              <Dialog.Panel className="w-full max-w-md transform overflow-hidden rounded-md bg-white p-6 text-left align-middle shadow-xl transition-all">
+                <Dialog.Title
+                  as="div"
+                  className="mb-2 text-lg font-semibold tracking-wide"
+                >
+                  {title}
+                </Dialog.Title>
+                {children}
+              </Dialog.Panel>
+            </Transition.Child>
+          </div>
+        </div>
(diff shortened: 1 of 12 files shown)


frontend2/src/views/MyTeam.tsx
```diff
@@ -143,6 +151,36 @@ const MyTeam: React.FC = () => {
...
+        <div className="mt-4 flex flex-col gap-2">
+          <p>
+            Are you sure you want to leave{" "}
+            <span className="font-semibold">{team.name}</span>?
+          </p>
+          <div className="flex flex-row gap-4">
+            <Button
+              variant="danger-outline"
+              onClick={() => {
+                void leaveMyTeam();
```
lowtorola, line 170: Should we await the leave team function here so the modal only closes once that request has resolved? With a spinner in the button?

frontend2/src/components/elements/Button.tsx
```diff
@@ -4,6 +4,9 @@ import Icon, { type IconName } from "./Icon";
 const VARIANTS: Record<string, string> = {
   "": "bg-gray-50 text-gray-800 hover:bg-gray-100 hover:ring-gray-900 hover:text-black ring-gray-500 ring-1 ring-inset",
   dark: "bg-cyan-700 text-white hover:bg-cyan-800",
+  danger: "bg-red-700 text-white hover:bg-red-800",
```
lowtorola, line 7: Nit: should we make a "type variant = 'dark' | 'danger' | ..." so this can be a record from variant to string?

frontend2/src/components/Modal.tsx
```diff
@@ -0,0 +1,60 @@
...
+import React, { Fragment } from "react";
+
+interface ModalProps {
+  isOpen: boolean;
+  closeModal: () => void;
+  title: string;
+  children?: React.ReactNode;
+}
+
+const Modal: React.FC<ModalProps> = ({
```
lowtorola, line 11: I remember we discussed this but why did we decide to name the dialog modal again? Can't remember but OK with it regardless 👍

lowtorola (approved):
Left a few comments but LGTM!

frontend2/src/components/elements/Button.tsx
lowtorola: This would be useful for type inference for example

frontend2/src/components/Modal.tsx
acrantel: My reasoning was so it wouldn't collide with the headless ui's naming of "Dialog" and because Popup might be confused with headless ui's popover

lowtorola: Ok I remember now SGTM!

frontend2/src/components/elements/Button.tsx
```diff
@@ -36,7 +44,15 @@ const Button: React.FC<ButtonProps> = ({
       {...rest}
     >
       {iconName !== undefined && <Icon name={iconName} size="sm" />}
-      {label}
+      {loading ? (
+        <Spinner
+          className="mx-2"
+          variant={variant.includes("danger") ? "danger" : ""}
```
lowtorola, line 50: Ahh I think this could be another use of variant typing... not sure this `.includes(...)` is very safe?

frontend2/src/contexts/CurrentTeamProvider.tsx
```diff
@@ -19,6 +19,11 @@ export const CurrentTeamProvider: React.FC<{ children: React.ReactNode }> = ({
   const { authState } = useCurrentUser();
   const { episodeId } = useEpisodeId();
 
+  const leaveMyTeam = useCallback(async (): Promise<void> => {
+    await leaveTeam(episodeId);
```
lowtorola, line 23: Should this callback have `episodeId` as a dependency?

lowtorola (approved):
Noticed a few minor things.. I'll approve so you can merge when you think it's good 🎉

frontend2/src/components/elements/Button.tsx
acrantel: I'll change it to === checks

frontend2/src/contexts/CurrentTeamProvider.tsx
acrantel: Yes! thanks for catching this

PR #697, opened 2023-11-04
Table Improvements

lowtorola:
- refactor all usages of BattlecodeTable.tsx into their own files in the components/ directory (and maybe rename BattlecodeTable to something less verbose :))
- Don't use idx's for keys in battlecodetable


acrantel (approved):
LGTM!

PR #698, opened 2023-11-12
Episode Switching

lowtorola:
Adds Episode Switching to frontend!

### DIFF
--- a/frontend2/src/components/EpisodeSwitcher.tsx
+++ b/frontend2/src/components/EpisodeSwitcher.tsx
@@ -1,0 +1,70 @@
+import React, { Fragment, useMemo } from "react";
+import { Listbox, Transition } from "@headlessui/react";
+import Icon from "./elements/Icon";
+
+interface EpisodeSwitcherProps<T extends React.Key | null | undefined> {
+  options: Array<{ value: T; label: string }>;
+  value: T;
+  onChange?: (value: T) => void;
+}
+
+function EpisodeSwitcher<T extends React.Key | null | undefined>({
+  options,
+  value,
+  onChange,
+}: EpisodeSwitcherProps<T>): JSX.Element {
+  const valueToLabel = useMemo(
+    () => new Map(options.map((option) => [option.value, option.label])),
+    [options],
+  );
+  return (
+    <div className={`relative`}>
+      <Listbox value={value} onChange={onChange}>
+        <div className="relative">
+          <Listbox.Button
+            className={`relative h-9 w-full truncate rounded-full bg-gray-900/80 py-1.5
+            pl-3.5 pr-8 text-left text-gray-200 shadow-sm focus:outline-none
+            sm:text-sm sm:leading-6`}
+          >
+            <span className="text-sm font-semibold">
+              {valueToLabel.get(value)}
+            </span>
+            <div
+              className="absolute inset-y-0 right-0 mr-2 flex transform items-center
+              transition duration-300 ui-open:rotate-180"
+            >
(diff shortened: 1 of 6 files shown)


frontend2/src/contexts/EpisodeContext.ts
```diff
@@ -27,3 +28,8 @@ export const useEpisodeId = (): EpisodeIdContextType => {
   }
   return episodeIdContext;
 };
+
+/** Use this function to retrieve all episodes. */
```
acrantel, line 32: I'd add a comment here saying that it is undefined if the data hasn't loaded yet

frontend2/src/components/EpisodeSwitcher.tsx
acrantel, line 1: @lowtorola you should review this file that I added

acrantel (approved):
LGTM! I added a file EpisodeSwitcher that should be reviewed as well

frontend2/src/components/EpisodeSwitcher.tsx
```diff
@@ -0,0 +1,75 @@
+import React, { Fragment, useMemo } from "react";
+import { Listbox, Transition } from "@headlessui/react";
+import Icon from "./elements/Icon";
+import { useEpisodeId, useEpisodeList } from "../contexts/EpisodeContext";
+import { isPresent } from "../utils/utilTypes";
+import { useNavigate } from "react-router-dom";
+
+const EpisodeSwitcher: React.FC = () => {
```
lowtorola, line 8: Looks like this file is the SelectMenu but styled/moved logic into from Header...why exactly did you choose to fully rewrite the SelectMenu skeleton here (as opposed to i.e. building a styled variant of SelectMenu)? Just want to make sure we're not adding boilerplate code 😎

frontend2/src/contexts/EpisodeContext.ts
lowtorola: Interesting... I think that this is pretty standard across our frontend APIs? Like inferrable from "Maybe"? If you think it's useful though ig we can add :+1:

lowtorola (commented):
Left some comments! This is huge though, enables dynamic episode markdown, home page, and more!

n8kim1:
holy heck let's go! thanks for the hard work, mega exciting

lowtorola:
> holy heck let's go! thanks for the hard work, mega exciting

The credit on this all goes to @acrantel! She did a great job with the App layout, all this constituted was adding a quick context and dropdown 🎉

frontend2/src/contexts/EpisodeContext.ts
acrantel: That's true - up to you then!

frontend2/src/components/EpisodeSwitcher.tsx
acrantel: I moved it out b/c EpisodeSwitcher doesn't seem like it belongs as a selectmenu variant and I wanted to avoid bloat in selectmenu
SelectMenu is something that should be used mainly in forms imo

frontend2/src/contexts/EpisodeContext.ts
lowtorola: Think I'll omit it... if someone gets confused I'll add it haha

frontend2/src/components/EpisodeSwitcher.tsx
lowtorola: Ok sounds good!

PR #699, opened 2023-11-17
add updateteampartial api call, add refreshTeam to team context

acrantel:
Add refreshTeam() to team context to enable updates of team state (needed for the Create/Join Team and My Team pages)
Add updateTeamPartial() api call (needed for My Team page)

will unblock @sri240 for create/join team

### DIFF
--- a/frontend2/src/contexts/CurrentTeamProvider.tsx
+++ b/frontend2/src/contexts/CurrentTeamProvider.tsx
@@ -24,6 +24,13 @@ export const CurrentTeamProvider: React.FC<{ children: React.ReactNode }> = ({
     setTeamData({ teamState: TeamStateEnum.NO_TEAM });
   }, [episodeId]);
 
+  const refreshTeam = useCallback((updatedTeam: TeamPrivate): void => {
+    setTeamData({
+      team: updatedTeam,
+      teamState: TeamStateEnum.IN_TEAM,
+    });
+  }, []);
+
   useEffect(() => {
     const loadTeam = async (): Promise<void> => {
       try {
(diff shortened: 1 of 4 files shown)


frontend2/src/utils/api/team.ts
```diff
@@ -63,6 +64,22 @@ export const retrieveTeam = async (episodeId: string): Promise<TeamPrivate> => {
...
 
+/**
+ * Push a partial update to the current user's team for a specific episode.
+ * @param episodeId The episode of the team
+ * @param teamProfilePrivateRequest The partial team update
+ * @returns The updated TeamPrivate object
+ */
+export const updateTeamPartial = async (
+  episodeId: string,
+  teamProfilePrivateRequest: TeamProfilePrivateRequest,
```
lowtorola, line 75: What is our standard practice for exposed API functions? I had thought we were listing parameters, i.e. `episodeId, teamId, userId, ...` instead of the whole `patchedTeamPrivateRequest`?

frontend2/src/contexts/CurrentTeamProvider.tsx
```diff
@@ -44,7 +51,9 @@ export const CurrentTeamProvider: React.FC<{ children: React.ReactNode }> = ({
   }, [authState, episodeId]);
 
   return (
-    <CurrentTeamContext.Provider value={{ ...teamData, leaveMyTeam }}>
+    <CurrentTeamContext.Provider
```
lowtorola, line 54: Looks like `refreshTeam` sets `teamState` to `IN_TEAM` no matter what? Is this only callable if the user is already in a team?

lowtorola (approved):
Sent a few comments, but otherwise looks good!

frontend2/src/utils/api/team.ts
acrantel: I think if there are a small number of parameters, we can list them. `TeamProfilePrivateRequest` is an object with a lot of fields though, so its easier to take in a `TeamProfilePrivateRequest` than all the fields

frontend2/src/contexts/CurrentTeamProvider.tsx
acrantel: Yeah, this function should only be called when the user is already on a team and has received an updated TeamPrivate object from some api call

acrantel: I'll add a comment describing this

PR #700, opened 2023-11-17
Support changing eligibility and scrim settings

acrantel:
Make the big checkboxes for changing eligibility / scrim settings work.
When reviewing / testing, please especially watch out for race conditions. I think I've addressed all potential race conditions, but would appreciate another check.

Testing plan:
* Be on a team. Click a checkbox. The spinner should appear in the checkbox until the request is resolved, and the checkbox should be set to the opposite of what it was before. 
* The checkbox should not be clickable while it is in the loading state.
* Clicking multiple checkboxes fast should not lead to inconsistent behavior

Merge after https://github.com/battlecode/galaxy/pull/699

Example: 
image

### DIFF
--- a/frontend2/src/components/team/EligibilitySettings.tsx
+++ b/frontend2/src/components/team/EligibilitySettings.tsx
@@ -10,6 +10,7 @@ import DescriptiveCheckbox, {
 import { useEpisode, useEpisodeId } from "../../contexts/EpisodeContext";
 import { isPresent } from "../../utils/utilTypes";
 import { updateTeamPartial } from "../../utils/api/team";
+import { isEqual } from "lodash";
 
 export function determineCheckboxState(
   inDesired: boolean,
@@ -40,9 +41,8 @@ const EligibilitySettings: React.FC = () => {
   useEffect(() => {
     // if actual team eligibility hasn't loaded yet or the desired/actual are identical
     if (
-      (!isPresent(desiredEligibility) ||
-        desiredEligibility === team?.profile?.eligible_for) ??
-      []
+      !isPresent(desiredEligibility) ||
+      isEqual(desiredEligibility, team?.profile?.eligible_for)
     ) {
       return;
     }
@@ -56,10 +56,12 @@ const EligibilitySettings: React.FC = () => {
       if (isActive && isPresent(newProfile) && isPresent(team)) {
         // only update the eligibility of the team profile to avoid race conditions
         // with other team profile updaters on this page
-        newProfile.eligible_for = updatedTeam.profile?.eligible_for;
         refreshTeam({
           ...team,
-          profile: newProfile,
+          profile: {
+            ...newProfile,
+            eligible_for: updatedTeam.profile?.eligible_for,
+          },
         });
       }
     };
@@ -74,7 +76,7 @@ const EligibilitySettings: React.FC = () => {
     !isPresent(team) ||
     !isPresent(episode)
   ) {
-    return null;
+    return "Error: You're not in a team!";
   }
   return (
     <SectionCard title="Eligibility">
--- a/frontend2/src/components/team/EligibilitySettings.tsx
+++ b/frontend2/src/components/team/EligibilitySettings.tsx
@@ -1,0 +1,126 @@
+import React, { useEffect, useState } from "react";
+import {
+  useCurrentTeam,
+  TeamStateEnum,
+} from "../../contexts/CurrentTeamContext";
+import SectionCard from "../SectionCard";
+import DescriptiveCheckbox, {
+  CheckboxState,
+} from "../elements/DescriptiveCheckbox";
+import { useEpisode, useEpisodeId } from "../../contexts/EpisodeContext";
+import { isPresent } from "../../utils/utilTypes";
+import { updateTeamPartial } from "../../utils/api/team";
+
+export function determineCheckboxState(
+  inDesired: boolean,
+  inActual: boolean,
+): CheckboxState {
+  if (inDesired && inActual) {
+    return CheckboxState.CHECKED;
+  } else if (!inDesired && !inActual) {
+    return CheckboxState.UNCHECKED;
+  } else {
+    return CheckboxState.LOADING;
+  }
+}
+
+// This component should only be used when there is a logged in user with a team.
+const EligibilitySettings: React.FC = () => {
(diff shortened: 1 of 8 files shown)


frontend2/src/components/team/EligibilitySettings.tsx
```diff
@@ -0,0 +1,126 @@
...
+  >();
+  useEffect(() => {
+    setDesiredEligibility(team?.profile?.eligible_for);
+  }, [team?.profile?.eligible_for]);
+
+  useEffect(() => {
+    // if actual team eligibility hasn't loaded yet or the desired/actual are identical
+    if (
+      (!isPresent(desiredEligibility) ||
+        desiredEligibility === team?.profile?.eligible_for) ??
```
acrantel, line 44: This check doesn't work because the two eligibility arrays aren't the same object. TODO: figure out how to compare these better .. or just remove the check for simplicity?

frontend2/src/components/sidebar/__test__/Sidebar.test.tsx
```diff
@@ -45,6 +45,8 @@ test("UI: should collapse sidebar", () => {
             leaveMyTeam: async (): Promise<void> => {
               await Promise.resolve();
             },
+            // eslint-disable-next-line @typescript-eslint/no-empty-function
+            refreshTeam: (team): void => {},
```
lowtorola, line 49: Ope I think you have this in two PRs haha... I would make it so it's only in one to save yourself a headache 😆

frontend2/src/components/team/EligibilitySettings.tsx
lowtorola: Consider `lodash.eq`?

lowtorola: [https://www.geeksforgeeks.org/lodash-_-eq-method/](https://www.geeksforgeeks.org/lodash-_-eq-method/)

frontend2/src/components/team/EligibilitySettings.tsx
```diff
@@ -0,0 +1,126 @@
...
+    let isActive = true;
+    const update = async (): Promise<void> => {
+      const updatedTeam = await updateTeamPartial(episodeId, {
+        eligible_for: desiredEligibility,
+      });
+      const newProfile = team?.profile;
+      if (isActive && isPresent(newProfile) && isPresent(team)) {
+        // only update the eligibility of the team profile to avoid race conditions
+        // with other team profile updaters on this page
+        newProfile.eligible_for = updatedTeam.profile?.eligible_for;
```
lowtorola, line 59: nit: do we prefer spreading the "new profile" in the refreshTeam call -> i.e. 
`refreshTeam({ 
    ...team, 
    profile: { ...newProfile, eligible_for: updatedTeam.profile?.eligible_for } 
});`

frontend2/src/components/team/EligibilitySettings.tsx
```diff
@@ -0,0 +1,126 @@
...
+      isActive = false;
+    };
+  }, [desiredEligibility, episodeId]);
+
+  if (
+    teamState !== TeamStateEnum.IN_TEAM ||
+    !isPresent(team) ||
+    !isPresent(episode)
+  ) {
+    return null;
```
lowtorola, line 77: Should we have a message here to say "you're not in a team!" or smth?

frontend2/src/components/team/EligibilitySettings.tsx
```diff
@@ -0,0 +1,126 @@
...
+              return (
+                <DescriptiveCheckbox
+                  key={criterion.id}
+                  status={determineCheckboxState(
+                    desiredEligibility.includes(criterion.id),
+                    (team.profile?.eligible_for ?? []).includes(criterion.id),
+                  )}
+                  onChange={(checked) => {
+                    if (checked && !desiredEligibility.includes(criterion.id)) {
+                      setDesiredEligibility([
```
lowtorola, line 103: nit: I think we should prefer spreading the previous state value, i.e. `setDesiredEligibility((prev) => [...prev, criterion.id])`.

frontend2/src/components/team/EligibilitySettings.tsx
```diff
@@ -0,0 +1,126 @@
...
+                        ...desiredEligibility,
+                        criterion.id,
+                      ]);
+                    } else if (!checked) {
+                      setDesiredEligibility(
```
lowtorola, line 108: ditto to above comment

frontend2/src/components/team/ScrimmageSettings.tsx
```diff
@@ -0,0 +1,119 @@
...
+      !isPresent(unranked) ||
+      (team?.profile?.auto_accept_ranked === ranked &&
+        team.profile.auto_accept_unranked === unranked)
+    ) {
+      return;
+    }
+
+    let isActive = true;
+    const update = async (): Promise<void> => {
+      const updatedTeam = await updateTeamPartial(episodeId, {
```
lowtorola, line 47: Will we get scary errors if we don't try/catch `await`s? I think maybe we should try to always try/catch them for safety?

lowtorola (approved):
Left some comments, but very exciting progress!

frontend2/src/components/team/EligibilitySettings.tsx
acrantel: yup! will change

frontend2/src/components/team/EligibilitySettings.tsx
acrantel: I can return a string that says that. 
I also have a comment on the component that says the current user should be in a team

frontend2/src/components/team/ScrimmageSettings.tsx
acrantel: I can add a console.error(error) here. In the future I think we should pop up a little notif box in a corner of the screen that say "request failed" or something

frontend2/src/components/team/EligibilitySettings.tsx
acrantel: This leads to some typing errors b/c typescript thinks that `prev` is `undefined || number[]`. 
(Doing it the original way works b/c I check desiredEligibility's definedness earlier in the function)

PR #703, opened 2023-11-19
Submission Uploads

lowtorola:
Will finish the Submissions Page!


frontend2/src/components/FileUpload.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+    return (
+      <div className={`relative ${invalid ? "mb-1" : ""} ${className}`}>
+        {label !== undefined && <FormLabel label={label} />}
+        <label className="flex max-w-md flex-row items-center">
+          <input
+            aria-invalid={errorMessage !== undefined ? "true" : "false"}
+            type="file"
+            className="hidden"
+            ref={ref}
```
lowtorola, line 46: I need this to be `fileUploadRef` so I can get the file name to display (the `forwardedRef` doesn't have this) :(

frontend2/src/views/Submissions.tsx
```diff
@@ -23,6 +31,21 @@ const Submissions: React.FC = () => {
   const [tourneySubData, setTourneySubData] =
     useState<Maybe<TournamentSubmission[]>>();
 
+  const [submitting, setSubmitting] = useState<boolean>(false);
+  const [subError, setSubError] = useState<Maybe<string>>();
+
+  const onSubmit: SubmitHandler<SubmissionUploadRequest> = async (data) => {
```
lowtorola, line 37: I never hit this function 🥲

frontend2/src/views/Submissions.tsx
```diff
@@ -91,6 +114,76 @@ const Submissions: React.FC = () => {
...
+          <p className="rounded-md border border-gray-400 bg-gray-100 px-4 py-3 font-mono">
+            {
+              "submission.zip --> examplefuncsplayer --> RobotPlayer.java, FooBar.java"
+            }
+          </p>
+        </span>
+        <form
+          // eslint-disable-next-line @typescript-eslint/no-misused-promises
+          onSubmit={(event) => {
+            event.preventDefault();
```
lowtorola, line 146: This is to stop the page from reloading (which is the default behavior).

frontend2/src/views/Submissions.tsx
```diff
@@ -91,6 +114,76 @@ const Submissions: React.FC = () => {
...
+          className="mt-4 flex flex-col"
+        >
+          <FileUpload
+            label="Choose Submission File"
+            fileInputHelp="Java .zip File (5MB max)"
+            errorMessage={subError}
+            required
+            {...register("file", { required: FIELD_REQUIRED_ERROR_MSG })}
+          />
+          {/* <div className="flex w-full flex-row items-center gap-10"> */}
```
lowtorola, line 158: Will these work if they're wrapped in a div for styling?

lowtorola:
@acrantel My form doesn't work but if it can be fixed then this page is done 👀

frontend2/src/views/Submissions.tsx
```diff
@@ -85,12 +115,139 @@ const Submissions: React.FC = () => {
...
+                  "submission.zip --> examplefuncsplayer --> RobotPlayer.java, FooBar.java"
+                }
+              </p>
+            </span>
+            <form
+              // eslint-disable-next-line @typescript-eslint/no-misused-promises
+              onSubmit={(event) => {
+                event.preventDefault();
+                handleSubmit(onSubmit);
+              }}
```
acrantel, line 210: Try changing this to 
```suggestion
              onSubmit={handleSubmit(onSubmit)}
```

frontend2/src/views/Submissions.tsx
acrantel: see my comment below about changing how handleSubmit is called

frontend2/src/views/Submissions.tsx
lowtorola: This causes the page to reload... not sure why?

frontend2/src/utils/api/compete.ts
```diff
@@ -21,19 +22,13 @@ const API = new CompeteApi(DEFAULT_API_CONFIGURATION);
  */
 export const uploadSubmission = async (
   episodeId: string,
-  submission: {
-    file: File;
-    packageName: string;
-    description: string;
-  },
+  submission: SubmissionUploadRequest,
 ): Promise<void> => {
```
acrantel, line 26: 1. We should have this return a Promise<Submission> which contains the submission received by our backend. 
2. we can then render the submission on our frontend in the submission history table. otherwise, the user will need to reload the page to see the new submission

The rerendering / storing of the updated submission is probably something that can be made easier with react-query, so I'm OK with waiting for that to be added before doing (2)

frontend2/src/components/tables/submissions/TourneySubTable.tsx
```diff
@@ -15,7 +15,9 @@ const TourneySubTable: React.FC<TourneySubTableProps> = ({ data, loading }) => {
     <Table
       data={data ?? []}
       loading={loading}
-      keyFromValue={(match) => match.id.toString()}
+      keyFromValue={(match) =>
+        match.tournament.toString() + match.id.toString()
+      }
```
acrantel, line 20: The recommendation below is so that the encoding is unambiguous (note that match ids should always be a number)
```suggestion
      keyFromValue={(match) =>
        match.tournament.toString() + "match" + match.id.toString()
      }
```

acrantel (commented):
This is great! Added a few comments

acrantel (approved):
lgtm! 👍

PR #707, opened 2023-11-30
React Query

lowtorola:
Adds React Query as `frontend2`'s async state manager! Plus, restructures our api!

New API structure:
- `src/api/`: folder for all api code
  - `_autogen`: ALL auto-generated code. Frontend code should only import models and types from this folder!
  - `"endpoint"`: Each API endpoint (e.g. episode, team, compete,...) gets its own sub-folder. Below is an example for episode to explain the layout.
    - `episodeApi.ts`: Wrapper instance methods for the `EpisodeApi` class. Basically just prettifying/safety enforcement through only exporting safe (non-"raw") functions.
    - `episodeKeys.ts`: `react-query` key factories for all Episode mutations and queries. These standardize the tags each api call is associated to in the `queryClient` cache, and simplifies query prefetching and invalidation.
    - `useEpisode.ts`: This holds all of our custom hooks that abstract away the `queryKeys` and other configuration parameters of `useQuery` and `useMutation`.


lowtorola:
Just noting that generated files are currently duplicated but that will be fixed once everything runs from `src/api` 😎

lowtorola:
Note to self, try this for queryFns to make it just work tm:

```
const defaultQueryFn = async ({ queryKey }) => {
  const { data } = await axios.get(
    `https://jsonplaceholder.typicode.com${queryKey[0]}`,
  )
  return data
}
```

frontend2/src/components/team/EligibilitySettings.tsx
```diff
@@ -100,18 +113,44 @@ const EligibilitySettings: React.FC = () => {
...
+                        episodeId,
+                        patchedTeamPrivateRequest: {
+                          profile: {
+                            ...team.profile,
+                            eligible_for: [
+                              ...(team.profile?.eligible_for ?? []),
+                              criterion.id,
+                            ],
+                          },
+                        },
```
acrantel, line 130: i think there will be a race condition here - an user can click multiple eligibilities in quick succession, but team.profile.eligible_for doersn't update until a request returns - thus the eligibilities will all override each other. 
I prevented this before by using desiredEligibility as a way to track the overall desired state on the client side

frontend2/src/components/team/EligibilitySettings.tsx
```diff
@@ -100,18 +113,44 @@ const EligibilitySettings: React.FC = () => {
...
-                        ...desiredEligibility,
-                        criterion.id,
-                      ]);
+                      // Add criterion to team
+                      // We do NOT want to await this, because we want to send the state of this mutation to isPending!
+                      updateTeam({
+                        episodeId,
+                        patchedTeamPrivateRequest: {
+                          profile: {
+                            ...team.profile,
```
acrantel, line 124: passing team.profile in here results in race conditions with the rest of the page (e.g. if ScrimmageSettings or MyTeam attempts to also update the profile, but their changes get overriden by this call)
we shouldn't need to pass team.profile here b/c it is a partial update

frontend2/src/components/team/ScrimmageSettings.tsx
```diff
@@ -32,52 +45,55 @@ const ScrimmageSettings: React.FC = () => {
...
+  //   };
+  //   void update();
+  //   return () => {
+  //     isActive = false;
+  //   };
+  // }, [ranked, unranked, episodeId]);
 
-  if (
-    teamState !== TeamStateEnum.IN_TEAM ||
+  // TODO: figure out a more elegant way to handle loading than blocking the UI
```
acrantel, line 92: The more elegant way would just be to put spaces like this where data is pending
image

frontend2/src/components/team/ScrimmageSettings.tsx
```diff
@@ -101,6 +117,16 @@ const ScrimmageSettings: React.FC = () => {
               team.profile?.auto_accept_ranked ?? false,
             )}
             onChange={(checked) => {
+              // TODO: does this object get passed in automatically? Can we just invoke the mutate function without passing in the object?
+              updateTeam({
+                episodeId,
+                patchedTeamPrivateRequest: {
+                  profile: {
+                    ...team.profile,
```
acrantel, line 125: we shouldn''t need these b/c it's a partial update, similar to the one in EligibiilitySettings
```suggestion
                 
```

lowtorola:
Huge progress on react querifying! Next step is fixing Eligibility/Scrim Settings sections. Then, I think I can tackle React Router loaders on this PR. Only problem is, I think that will make it necessary to refactor the `queryKey/queryFn` pairs for each hook to reuse on the loader functions. Also, I think I will refactor a `prefetchNextPage` function that many of our `use<Model>List` hooks use!

lowtorola:
@acrantel Progress update: I finished reworking `EligibilitySettings`, `ScrimmageSettings`, and `MyTeam` as well as implementing the useTeam and useCompete hooks across the FE. Once the useUser hooks are in from your side I think we can do a mega-test on this and merge!! 🎉

frontend2/src/components/team/ScrimmageSettings.tsx
lowtorola: TODO by someone sometime in the future... not now 😃

frontend2/src/components/elements/Button.tsx
```diff
@@ -30,15 +33,17 @@ const Button: React.FC<ButtonProps> = ({
   iconName,
   fullWidth = false,
   className = "",
+  disabled,
   ...rest
 }) => {
-  const variantStyle = `${VARIANTS[variant]}  ${
+  const variantStyle = `${disabled ? DISABLED : VARIANTS[variant]} ${
```
lowtorola, line 39: @acrantel Is this disabled implementation ok? Specifically my styling and disabling on load 😎

frontend2/src/components/elements/Button.tsx
```diff
@@ -28,12 +28,12 @@ interface ButtonProps extends React.ComponentPropsWithoutRef<"button"> {
 
 const Button: React.FC<ButtonProps> = ({
   variant = "",
-  disabled = false,
   loading = false,
   label,
   iconName,
   fullWidth = false,
   className = "",
+  disabled,
```
acrantel, line 36: oops it looks like your merge commit removed the `disabled=false` here which is necessary for eslint 
```suggestion
  disabled = false,
```

frontend2/src/components/elements/Button.tsx
acrantel: The disabling on load is great! We should have different variants of disabled buttons b/c some places expect dark buttons and a light disabled button looks strange (e.g. login page)
image

lowtorola:
Going to yeet the test files that are breaking CI 😢 maybe we will add real unit tests someday lol

frontend2/src/views/Login.tsx
```diff
@@ -32,13 +35,15 @@ const Login: React.FC = () => {
...
-      setLoginError(
-        "Error logging in. Did you enter your username and password correctly?",
-      );
+      await toast.promise(login(data.username, data.password, queryClient), {
+        loading: "Logging in...",
+        success: "Logged in!",
+        error:
+          "Error logging in. Did you enter your username and password correctly?",
+      });
+      navigate(episodeId !== undefined ? `/${episodeId}/home` : "/");
```
acrantel: I don't think we need this extra navigate (see the useEffect above that should accomplish the same thing)

lowtorola: 💯

lowtorola:
WHOOOO 🎉 🎉 🎉

PR #712, opened 2023-12-20
expose tournament round display order field

acrantel:
`display_order` specifies what order to display the `TournamentRounds` of a `Tournament`

Test plan
* ensure display_order is in the response of tournament round endpoints


lowtorola (approved):
LGTM! :tada:

PR #714, opened 2024-01-07
prep for 2024 release

acrantel:
Add scaffold link / name to frontend.
+ small update to client instructions in getting started

@ client please review the client instructions in getting started

image


TheApplePieGod:
Looks good to me, although I was doing some testing and I'm not sure about the accuracy of the jdk stuff. For one, the release workflow uses one called 'adopt,' which isn't oracle. Hank is also using a non oracle jdk during dev

TheApplePieGod (approved):

PR #719, opened 2024-01-10
fix multiepisode strategy report bug

acrantel:
Fixes a bug in the team report endpoints where we didn't filter for episode

Tested locally with a user on 2 teams. `GET /api/team/{episode_id}/requirement/report/` returned 404 which is expected (previously returned 500)


lowtorola (approved):
LGTM!

PR #723, opened 2024-01-15
Upgrade to use Challonge API v2.1

n8kim1:

PR #724, opened 2024-01-15
Better options for creating challonge tournaments

n8kim1:
]Attachments change: To actually _let_ us add attachments (in our case, replay links). Attaching via API doesn't work but at least you can do manually
Game name: This is required not by the API v2 but by the tournament settings _website_. (Either they do frontend validation or use v1 of API.) So if you ever change the settings manually via website, you have to go back and add the game name, which gets really painful.... Also all our other tournaments have game name anyways, so might as well keep that around.

Also includes some dev notes

### DIFF
--- a/backend/siarnaq/bracket/challonge.py
+++ b/backend/siarnaq/bracket/challonge.py
@@ -59,6 +59,7 @@ def create_tournament(tournament: Tournament, *, is_private: bool):
                 "tournament_type": challonge_type,
                 "private": is_private,
                 "url": challonge_id,
+                "game_name": "Battlecode",
                 "match_options": {"accept_attachments": True},
             },
         }
--- a/backend/siarnaq/bracket/challonge.py
+++ b/backend/siarnaq/bracket/challonge.py
@@ -60,7 +60,9 @@ def create_tournament(tournament: Tournament, *, is_private: bool):
                 "private": is_private,
                 "url": challonge_id,
                 "game_name": "Battlecode",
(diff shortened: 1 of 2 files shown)


backend/siarnaq/bracket/challonge.py
```diff
@@ -59,6 +59,8 @@ def create_tournament(tournament: Tournament, *, is_private: bool):
                 "tournament_type": challonge_type,
                 "private": is_private,
                 "url": challonge_id,
+                "game_name": "Battlecode",
```
acrantel, line 62: These changes just set up our tournaments to support replay url attachments in the future, correct?

acrantel (dismissed):
lgtm, great documentation!

backend/siarnaq/bracket/challonge.py
acrantel: Also, will we need to merge #723 before merging this PR? b/c I see changes from camelcase to snakecase in that PR, but i'm not sure if that's a challonge v2 -> v2.1 change or challonge v1 -> v2  change

n8kim1: Yup (and also they do more things)
 updated PR description for more context

n8kim1: ah good catch, idk if "game_name" is v2 or v2.1
I'll play it safe

nour-massri (dismissed):

Issue #605, opened 2023-01-24
Consider autoscaling tweaks for ends-of-tour-rounds to run better


n8kim1:
https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/compute_autoscaler#nested_scale_down_control
https://cloud.google.com/compute/docs/autoscaler#scale-in_controls
https://cloud.google.com/compute/docs/autoscaler/understanding-autoscaler-decisions#delays_in_scaling_in

n8kim1:
we did some stuff, so tis issue might be fixed. id check

PR #729, opened 2024-01-20
Making scaledown of match / execute servers more gradual and slower

n8kim1:
To alleviate #605 

This would slightly increase costs of year-round runs, especially when someone runs only one match in the random middle of the year. But those increases shouldn't be much anyways

Happy to tweak params, or just not do this anyways. I mainly made this PR just to close my tabs xp

https://cloud.google.com/compute/docs/autoscaler#scale-in_controls
https://cloud.google.com/compute/docs/autoscaler/understanding-autoscaler-decisions#delays_in_scaling_in

### DIFF
--- a/deploy/saturn/main.tf
+++ b/deploy/saturn/main.tf
@@ -161,6 +161,12 @@ resource "google_compute_autoscaler" "this" {
     max_replicas    = var.max_instances
     min_replicas    = var.min_instances
     cooldown_period = 60
+    scale_down_control {
+      max_scaled_down_replicas {
+         fixed =  1
+      }
+      time_window_sec =  600
+    }
 
     metric {
       name                       = "pubsub.googleapis.com/subscription/num_undelivered_messages"


j-mao:
Noting that #720 already alleviates this a lot by making there be less machines. If we can come up with a test plan to evaluate this before/after, we can deploy and evaluate what works better.

Noting also that scaling is already delayed by 1-2 mins because pub/sub metrics take time to propagate. So by the time scaling in happens, the scrim queue is actually far less than the assignment ratio.

n8kim1:
about #720 -- Yes good catch (had thought about that but forgot to mention). For a plan... what if I watch the queue during the next tournament as-is, and we can evaluate how much scrimmage servers interrupted? (like how much the bad behavior is still present)
I can also deploy this version and watch a tournament too

about delay -- good to know, thanks; That should help a ton for scaling in, and I can make scaling _out_ a bit faster. Unfortunately relying on the baked-in 1-2 minute delay isn't probably long enough on its own though

PR #730, opened 2024-01-22
log tracebacks for tasks, log errors properly

acrantel:
If an error occurs while running a task, log a traceback.

This PR also fixes the bug where the error from running a task was overriden/deleted by the error from calling FinalizeReport, resulting in our logs not containing the error.

**Testing**: not sure how to test this locally


lowtorola (approved):
:shipit:

PR #735, opened 2024-02-02
only allow scrimmages against reference player (revert later)

acrantel:
tested with reference player team in staging


lowtorola (approved):
Looks great assuming that it has been tested 🎉

PR #740, opened 2024-02-02
Home Page

lowtorola:
Adds a home page for frontend2. Currently just a slightly ugly skeleton but with functional dynamic layout! Think this is blocked by @sri240's ranking chart work but I think this is good to have so we can throw things in here.

Screenshot 2024-02-02 at 1 48 36 PM


acrantel (approved):
looks good to merge in! we can add sri's ranking chart in another PR when that's done

PR #761, opened 2024-03-14
Allow editing of fields for user page

nicolexxuu:
All fields should be editable now; both avatar and resume upload should work

* TODO: add resume download link so that submitted resume is viewable


backend/siarnaq/api/user/serializers.py
```diff
@@ -34,7 +34,7 @@ def update(self, *args, **kwargs):
 class UserProfilePrivateSerializer(UserProfilePublicSerializer):
     # Couuntry field requires special serialization.
```
n8kim1, line 35: nit: typo

lowtorola (approved):
@nicolexxuu LGTM! Feel free to hit merge! :shipit:

PR #772, opened 2024-04-06
Diego/charting

Desperationis:
Used the highcharts library to copy the chart in https://jmerle.github.io/battlecode-2024-statistics/ . This is not a 1:1 chart because they used a theme that did everything for them, while I opted to copy it using explicit settings. I put it in the home screen of /home:

image

The reason the data looks smooth is because its fake data I generated that has many thousand data points. If used on a lot less data points, it looks blocks like the one on the website. I still need to implement fetching data from the API


frontend2/src/components/tables/chart/TeamChart.tsx
```diff
@@ -0,0 +1,100 @@
...
+import HighchartsAccessibility from "highcharts/modules/accessibility";
+import * as chart from "./chart";
+import * as random_data from "./random_data";
+
+type UTCMilliTimestamp = number;
+
+export interface TeamChartProps {
+  yAxisLabel: string;
+
+  values: Record<string, Array<[UTCMilliTimestamp, number]>>;
```
acrantel, line 11: coudl be helpful to add a comment here explaining what the structure of values is (e.g. maps team names to a list of timestamps and ranking at that timestamp)

acrantel (approved):
lgtm! great job :)

PR #780, opened 2024-04-27
Rating History API

lowtorola:
Adds a rating history API for multiple teams at once! Also, connects the home page chart to said api.


lowtorola:
@Desperationis The home page chart looks a little odd (you can see if you pull this branch down). Also, I think we should implement a "loading" state for the chart component! Think you could take those tasks on? 😎

Desperationis:
I added a loading status for the chart. It's not the spiny thing you wanted, but you can customize the text and style to whatever you want. Also, the reason the chart looks weird on the home page is because, by the looks of it, there is no data being returned for that episode. If you change the episode to a different year (2023 / 2022) it works just fine. I added a "No data available" status for this

Desperationis:
Also, there is a really weird error and I have absolutely no clue where it is coming from. It doesn't stop the program from running, but there's a type error in console

lowtorola:
> Also, there is a really weird error and I have absolutely no clue where it is coming from. It doesn't stop the program from running, but there's a type error in console

Couldn't figure this out... just squashed it with a try catch haha 😆

frontend2/src/components/tables/chart/TeamChart.tsx
```diff
@@ -14,28 +19,62 @@ type UTCMilliTimestamp = number;
...
+    }));
+  }, [values]);
+
+  if (myChart !== undefined) {
+    try {
+      if (loading) myChart.showLoading(loadingMessage ?? "Loading...");
+      else myChart.hideLoading();
+    } catch (e) {
+      // Ignore internal highcharts errors...
+    }
```
acrantel, line 55: I think the reason there is an error here is because myChart.options is not defined. Could we define it before we call showLoading?

image

acrantel:
How are we testing this? There seems to be ranking data in the staging database for Battlecode 2022 but it is not showing up in the chart.

frontend2/src/components/tables/chart/TeamChart.tsx
acrantel: Ok trying to do that causes other errors nvm 💀

acrantel: actually i think we may just need to put it in a useeffect. 
image

lowtorola:
> How are we testing this? There seems to be ranking data in the staging database for Battlecode 2022 but it is not showing up in the chart.

@acrantel I see this in my staging... did you mean that there is more data that doesn't show up? Or like is there just none for you?

Screenshot 2024-05-01 at 11 19 23 PM

acrantel:
NVM disregard what I was saying about no data, I was testing it on staging backend when I should have been testing with local backend 💀 
It appears for me now!

acrantel (approved):
lgtm! this is a breaking change for our API, but I don't think our old frontend uses the historical_rating endpoint so it is nbd

PR #797, opened 2024-05-01
Routing Improvements

### LINKED ISSUE #794: Routing Lockdown
Currently, navigating to pages that don't match our pattern for navigation bring up the homepage and potentially throw a server error toast (because we are trying to parse an episode that doesn't exist). The following rules should be implemented:
 - [x] Episode info must be loaded before the home page renders 
  - [x] Await load episode info in the episodeLoader
  - [x] Indicate that the episode is loading in the episode switcher while it verifies that the episode exists
 - [x] If the episode info load fails, redirect to 404 not found page
 - [x] If the page load fails, redirect to 404 not found page
 - [x] Navigating to "localhost:3000" redirects to current episode home page

lowtorola:
Fixes some routing insecurities and improves flow of control around episode switching.
 - [x] Episode info must be loaded before the home page renders 
  - [x] Await load episode info in the episodeLoader
  - [x] Indicate that the episode is loading in the episode switcher while it verifies that the episode exists
 - [x] If the episode info load fails, redirect to 404 not found page
 - [x] If the page load fails, redirect to 404 not found page
 - [x] Navigating to "localhost:3000" redirects to current episode home page

Closes #794 :rocket:

### DIFF
--- a/frontend2/src/App.tsx
+++ b/frontend2/src/App.tsx
@@ -20,7 +20,7 @@ import {
   type LoaderFunction,
 } from "react-router-dom";
 import { DEFAULT_EPISODE } from "./utils/constants";
-import NotFound from "./views/NotFound";
+import EpisodeNotFound from "./views/EpisodeNotFound";
 import Rankings from "./views/Rankings";
 import { CurrentUserProvider } from "./contexts/CurrentUserProvider";
 import PrivateRoute from "./components/PrivateRoute";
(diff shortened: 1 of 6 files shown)


acrantel:
just to clarify - does "redirect to 404 not found page" mean that we should navigate to something like /notfound or does it mean that we will display the "page not found" message on the screen? currently it seems to be the "page not found" message.

(if the current behavior is the expected behavior, then this PR lgtm!)

lowtorola:
@acrantel Yes it navigates to the "Page Not Found" message screen. I thought that seemed logical since technically for those bad URLs they are trying to access a page that doesn't exist

acrantel (approved):
image

great work!

lowtorola:
:shipit:

PR #837, opened 2024-10-12
Safe Loaders

lowtorola:
This fixes the issue of loaders throwing runtime errors. I added a few notes, feel free to give your thoughts on them. tyty

### DIFF
--- a/frontend2/src/api/helpers.ts
+++ b/frontend2/src/api/helpers.ts
@@ -1,9 +1,11 @@
 import Cookies from "js-cookie";
 import { Configuration, type ResponseError } from "./_autogen";
 import type {
+  PaginatedQueryFactory,
   PaginatedQueryFuncBuilder,
   PaginatedRequestMinimal,
   PaginatedResultMinimal,
+  QueryFactory,
   QueryKeyBuilder,
 } from "./apiTypes";
 import type { QueryClient, QueryKey } from "@tanstack/react-query";
@@ -83,4 +85,37 @@ export const prefetchNextPage = async <
       toast.error((e as ResponseError).message);
     }
   }
+};
+
+const safeEnsureQueryDataHelper = async <T, K>(
+  request: T,
+  factory: QueryFactory<T, K> | PaginatedQueryFactory<T, K>,
+  queryClient: QueryClient,
+): Promise<void> => {
+  try {
+    await queryClient.ensureQueryData({
+      queryKey: buildKey(factory.queryKey, request),
+      // TypeScript allows passing in extra params so this works for regular & paginated
+      queryFn: async () => await factory.queryFn(request, queryClient, true),
+    });
+  } catch (err) {
+    // This error will have toasted from the QueryClient if necessary
+  }
+};
+
+/**
+ * Given a request and query identification, safely ensure that the query data exists
+ * or is fetched with no risk of throwing a Runtime Error. If the request has a server
+ * error, it will be displayed as a toast.
+ * @param request the parameters of the request
+ * @param queryKey
+ * @param queryFn
+ * @param queryClient
+ */
+export const safeEnsureQueryData = <T, K>(
+  request: T,
+  factory: QueryFactory<T, K> | PaginatedQueryFactory<T, K>,
+  queryClient: QueryClient,
+): void => {
+  void safeEnsureQueryDataHelper(request, factory, queryClient);
 };
--- a/frontend2/src/api/loaders/episodeLoader.ts
+++ b/frontend2/src/api/loaders/episodeLoader.ts
@@ -1,0 +1,47 @@
+import { type QueryClient } from "@tanstack/react-query";
+import { ResponseError } from "api/_autogen";
+import { loginCheck } from "api/auth/authApi";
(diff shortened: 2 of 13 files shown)


frontend2/src/api/helpers.ts
```diff
@@ -84,3 +86,42 @@ export const prefetchNextPage = async <
...
+
+const safeLoaderHelper = async <T, K>(
+  request: T,
+  factory: QueryFactory<T, K> | PaginatedQueryFactory<T, K>,
+  queryClient: QueryClient,
+): Promise<void> => {
+  try {
+    await queryClient.ensureQueryData({
+      queryKey: buildKey(factory.queryKey, request),
+      queryFn: async () => await factory.queryFn(request, queryClient, true),
```
lowtorola, line 98: I think there is some TS trickery going on here... this function signature matches `PaginatedQueryFuncBuilder` but works when called on that or `QueryFuncBuilder` ??? No idea how but it seems to work flawlessly.

frontend2/src/api/helpers.ts
```diff
@@ -84,3 +86,42 @@ export const prefetchNextPage = async <
...
+ * @param queryFn
+ * @param queryClient
+ */
+export const safeLoader = <T, K>(
+  request: T,
+  factory: QueryFactory<T, K> | PaginatedQueryFactory<T, K>,
+  queryClient: QueryClient,
+): void => {
+  safeLoaderHelper(request, factory, queryClient).catch((error) => {
+    let _ = 0;
```
lowtorola, line 122: Know of any better ways to trick the linter into not complaining that we're not "handling the error", even though the error handling is upstream? Maybe an eslint disable?

frontend2/src/api/loaders/episodeLoader.ts
```diff
@@ -0,0 +1,47 @@
...
+      throw new ResponseError(
+        new Response("Episode not found.", { status: 404 }),
+      );
+    }
+
+    // Run a check to see if the user has an invalid token
+    const loggedIn = await loginCheck(queryClient);
+
+    // Await the episode info so we can be sure that it exists
+    const episodeInfo = await queryClient.ensureQueryData({
```
lowtorola, line 25: I'm still awaiting things that I want to e.g. wait or throw to error boundary on :sunglasses:

frontend2/src/api/loaders/homeIfLoggedIn.ts
```diff
@@ -3,7 +3,7 @@ import { loginCheck } from "api/auth/authApi";
 import { type LoaderFunction } from "react-router-dom";
 import { DEFAULT_EPISODE } from "utils/constants";
 
-export const passwordForgotLoader =
+export const homeIfLoggedIn =
```
lowtorola, line 6: this is because I changed this to be used for login as well... know of any other pages that could use it?

frontend2/src/api/helpers.ts
acrantel: > In JavaScript, if you call a function with more arguments than there are parameters, the extra arguments are simply ignored. TypeScript behaves the same way. Functions with fewer parameters (of the same types) can always take the place of functions with more parameters.

https://www.typescriptlang.org/docs/handbook/2/functions.html

frontend2/src/api/helpers.ts
acrantel: we can put `void` in front of safeLoaderHelper

frontend2/src/api/helpers.ts
```diff
@@ -84,3 +86,42 @@ export const prefetchNextPage = async <
...
+  factory: QueryFactory<T, K> | PaginatedQueryFactory<T, K>,
+  queryClient: QueryClient,
+): Promise<void> => {
+  try {
+    await queryClient.ensureQueryData({
+      queryKey: buildKey(factory.queryKey, request),
+      queryFn: async () => await factory.queryFn(request, queryClient, true),
+    });
+  } catch (err) {
+    if ((err as ResponseError).response.status > 500) {
```
acrantel, line 101: `(err as ResponseError).response.status` errors if `err.response` is undefined, which can happen when err is not a ResponseError

frontend2/src/api/loaders/homeIfLoggedIn.ts
acrantel: registration page? not sure if there's anything else

frontend2/src/api/helpers.ts
```diff
@@ -84,3 +86,42 @@ export const prefetchNextPage = async <
...
+/**
+ * Given a request and query identification, safely ensure that the query data exists
+ * or is fetched with no risk of throwing a Runtime Error. If the request has a server
+ * error, it will be displayed as a toast.
+ * @param request the parameters of the request
+ * @param queryKey
+ * @param queryFn
+ * @param queryClient
+ */
+export const safeLoader = <T, K>(
```
acrantel, line 116: maybe we can rename these functions to something like safeEnsureQueryData

acrantel (approved):
lgtm! :shipit:

PR #849, opened 2024-11-17
Make user profile page (without ranking chart)

### LINKED ISSUE #787: Other User Page UI
Need to build out a UI for the "Other User" page. Can/should probably take strong inspiration from the "Other team" page, including the ranking graph, etc. I think it should be doable to do this in one shot since all of the UI should be read-based.

### LINKED ISSUE #845: Other User Page UI Buildout
The "other user" page has an existing UI skeleton (rather, it will after #787 is completed), but it needs to be connected to the API for more functionality and beautifulness! Here is some information about the task:

- The page has the path "/<episodeId>/user/<userId>"
-  Called UserProfile.tsx
-  It should be very similar to Account.tsx, but should have a biography (only the fields that are public!) and a chart that shows their rating history across past years of battlecode (Account has this chart)

 - start by reading the account and myteam pages. What will you need to copy over from them?
 - plan out what API hooks you will need to fetch the information. Look at pages like Home.tsx, Account.tsx, etc. (ask if you get stuck here)
 - begin adding components or fleshing them out! I always recommend stealing any classnames you might need from other places around the codebase
 - profit??? :shipit:

acrantel:
image

the path is `/user/<userid>` for these user pages

CLOSES #787 closes #845


frontend2/src/views/UserProfile.tsx
```diff
@@ -0,0 +1,92 @@
...
+                <div>
+                  <div className="font-semibold">Team</div>
+                  {isNilOrEmptyStr(userTeam.data?.[episodeId]?.name) ? (
+                    <span className="italic text-gray-500">
+                      This user does not have a team.
+                    </span>
+                  ) : (
+                    <Link
+                      to={`/${episodeId}/team/${
+                        userTeam.data?.[episodeId]?.id ?? 0
```
lowtorola, line 50: nit: y is episodeId in square brackets? not hating on the syntax just haven't seen that before

frontend2/src/views/UserProfile.tsx
```diff
@@ -0,0 +1,92 @@
...
+            </div>
+          </SectionCard>
+          <SectionCard title="Rating History"></SectionCard>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+const UserNotFound: React.FC = () => {
```
lowtorola, line 87: I will change this behavior in #854 yippeeee

t
lowtorola, line 1: what is this? :smiley:

frontend2/src/views/UserProfile.tsx
acrantel: userTeam.data is a record

t
acrantel: deleting it idk what it is

lowtorola (dismissed):
LGTM! :shipit:

lowtorola (approved):
LGTM! :shipit:

PR #853, opened 2024-11-22
update sidebar for episode not released, add errors for episode released-only pages

lowtorola:
adds some checks to make sure that users can't access pages we don't want them to :sunglasses: 

* refactors sidebar logic, using an enum for each sidebar item to describe when it should be displayed

### DIFF
--- a/frontend2/src/api/loaders/rankingsLoader.ts
+++ b/frontend2/src/api/loaders/rankingsLoader.ts
@@ -1,15 +1,30 @@
 import type { QueryClient } from "@tanstack/react-query";
-import type { LoaderFunction } from "react-router-dom";
+import { redirect, type LoaderFunction } from "react-router-dom";
 import { episodeInfoFactory } from "../episode/episodeFactories";
-import { safeEnsureQueryData } from "../helpers";
+import { buildKey, safeEnsureQueryData } from "../helpers";
 import { searchTeamsFactory } from "../team/teamFactories";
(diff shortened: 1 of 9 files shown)


frontend2/src/components/Header.tsx
```diff
@@ -1,19 +1,30 @@
...
   const { episodeId } = useEpisodeId();
   const queryClient = useQueryClient();
+
+  const teamData = useUserTeam({ episodeId });
+  const episodeData = useEpisodeInfo({ id: episodeId });
+
+  const renderedItems = useMemo(
+    () => renderableItems(ALL_SIDEBAR_ITEMS, episodeData, authState, teamData),
+    [episodeData, teamData],
+  );
```
acrantel, line 26: note that this is repeated from sidebarsection but i think it's fine

frontend2/src/api/loaders/rankingsLoader.ts
```diff
@@ -1,16 +1,31 @@
...
+    });
+
+    if ((await episodeData).game_release.getTime() > Date.now()) {
+      toast.error(
+        `Rankings page not released yet for ${(await episodeData).name_long}.`,
+      );
+      return redirect(`/${episodeId}/home`);
+    }
+
     // Episode info
```
acrantel, line 29: we should make ranking page avaialble pre release so people can see teams, also make it visible in sidebar

PR #857, opened 2024-12-06
Fix content page errors and scrimmaging record error

### LINKED ISSUE #855: runtime error when episode doesn't have content defined
There is a runtime error in frontend2 when an episode doesn't have content defined for the resources, quickstart, tournament pages (and probably any other page that uses content from the src/content folder. Reproduce by going to pages such as http://localhost:3000/bc24test/tournaments when running local frontend2 with staging backend. 

Instead of runtime errorring, let's default to something like "resources have not been published for this episode" when content doesn't exist for an episode?

lowtorola:
- closes #855 :rocket:
- fixes annoying toast error caused by scrim record endpoint


frontend2/src/views/Tournaments.tsx
```diff
@@ -33,12 +34,14 @@ const Tournaments: React.FC = () => {
...
-  const hasContent = Object.values(currentTournamentText).some(
-    (value) => value !== "",
-  );
+  const hasContent =
+    !isNil(currentTournamentText) &&
+    Object.values(currentTournamentText).some((value) => value !== "");
+
   if (!hasContent) {
-    return <NoContentLegacyEpisode />;
+    return <NoContentFound />;
```
acrantel, line 44: so to display the tournaments that are registered in a released episode, this code now requires that content is specified for the tournaments page. i think this is fine.

acrantel (approved):
lgtm, don't forget to release backend!

PR #864, opened 2024-12-10
Create content for battlecode 2025

lowtorola:
- content for new episodes
- remove content for staging-only bc25


nicolexxuu (approved):
so true 👍🏻

PR #873, opened 2025-01-03
Add python3 and java21 support on saturn

lowtorola:
- adds language options for java21
- adds ability for saturn runners to run matches with java21 and python3

### DIFF
--- a/saturn/Dockerfile
+++ b/saturn/Dockerfile
@@ -10,7 +10,16 @@ COPY . .
 RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o /saturn -ldflags="-s -w" ./cmd/saturn/main.go
 
 
-FROM openjdk:8-slim-buster
+FROM openjdk:21-jdk-slim-buster
+
+# Install JDK8
+RUN apt-get update; apt-get install -y openjdk-8-jdk
+
+# Install python
+RUN apt-get install -y software-properties-common && \
+    add-apt-repository -y ppa:deadsnakes/ppa && \
+    apt-get update && \
+    apt install -y python3.12
 
 ENV APP_HOME /app
 WORKDIR $APP_HOME
--- a/saturn/Dockerfile
+++ b/saturn/Dockerfile
@@ -1,4 +1,4 @@
-FROM golang:1.18-buster AS go
+FROM golang:1.20.1-buster AS go
 
 ENV BUILD_HOME /build
 WORKDIR $BUILD_HOME
--- a/saturn/Dockerfile
+++ b/saturn/Dockerfile
@@ -14,10 +14,12 @@ FROM openjdk:21-jdk-slim-bookworm
 
 # Install JDK8. The base image provides JDK21, but we still need JDK8 to
 # run matches with java8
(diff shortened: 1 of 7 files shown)


saturn/Dockerfile
```diff
@@ -10,7 +10,16 @@ COPY . .
 RUN GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o /saturn -ldflags="-s -w" ./cmd/saturn/main.go
 
 
-FROM openjdk:8-slim-buster
+FROM openjdk:21-jdk-slim-buster
```
lowtorola, line 13: So the idea is that now saturn will only ever use java21? as opposed to running legacy matches with java8? I would like a clarification on why exactly we are doing this if so

saturn/Dockerfile
```diff
@@ -10,7 +10,16 @@ COPY . .
...
+
+# Install JDK8
+RUN apt-get update; apt-get install -y openjdk-8-jdk
```
lowtorola, line 16: should we have a matching install for java21?

saturn/pkg/run/java8.go
lowtorola, line 1: do we need a matching `java21.go`?

lowtorola: yeah after reading the file i suggest creating java21.go unless there's a really good reason not to

saturn/pkg/run/scaffold.go
```diff
@@ -115,13 +116,24 @@ type Scaffold struct {
...
+		s, err := NewJava8Scaffold(ctx, episode, repo, root, "/usr/lib/jvm/java-8-openjdk-amd64")
 		if err != nil {
 			return nil, fmt.Errorf("NewJava8Scaffold: %v", err)
 		}
 		return &s.Scaffold, nil
+	case saturn.Java21:
+        // Java21 uses the same scaffold for now. Modern java21 scaffolds store
+        // java in the 'java' subdirectory of the scaffold
+        javaRoot := filepath.Join(root, "java")
+		s, err := NewJava8Scaffold(ctx, episode, repo, javaRoot, "/usr/local/openjdk-21")
```
lowtorola, line 129: ok i see the vision now... i think I'd prefer separate scaffolds even though it should be kinda copy/pasta

lowtorola (commented):
Needs the new language option added to backend which i will do on my local... funny that since i created the PR I can't approve.

See my comments about separate scaffold files for java21/java8. I think that is a preferable paradigm to reusing the java8 scaffold for java21 but im open to discussion.

saturn/Dockerfile
TheApplePieGod: We are just using 21 as the base image, we install jdk8 separately so we can still use it

saturn/Dockerfile
TheApplePieGod: See above

saturn/Dockerfile
lowtorola: pls add comment explaining tyty

TheApplePieGod (approved):

TheApplePieGod (approved):

PR #879, opened 2025-01-06
Fix delta

erikji:
Bug: UI shows both teams as gaining rating, even though one team lost rating
Screenshot 2025-01-05 at 21 38 45

Also fixed double negative sign


lowtorola (approved):
LGTM! ty for the assist @erikji :rocket:

PR #900, opened 2025-01-10
Remove duplicated team biography on team profile view

spsquared:
Removes "Team Biography" section below "Biography" on team profile page. ("Biography" more consistent with rest of profile than "Team Biography")


lowtorola (approved):
LGTM! thanks for contributing!

PR #903, opened 2025-01-13
Tournament admin page

lowtorola:
this is like halfway done? things that are sus rn:
- when exactly can we create/enqueue?
  - TODO actually refetch rounds on c/e success :)
  - TODO better error messages when things fail
  - active round is... the earliest round that is "in progress"
  - we can rerun the round once all matches are completed only if the next round isnt in progress. this should rerun all the matches using match "requeue" functionality (need a new api request on tournament round which forcibly requeues all its matches)
  - you can only start a round once the previous round's matches are all successful/completed
  - 
- when can we release? what exactly does releasing do?
  - we dont release until the next day (until stream)
  - as the rounds are shown we release them
  - rounds can be released once all their matches are succesful (unimportant feature)
- what does releasestatusenum mean?
  - 
- how to link to challonge?
  - tourney has two ids, use challonge.com/"external private id"
- why aren't maps showing up in the tournament round?
  - TODO lowell debug
- do we actually need more than 10 round tabs to show (if so, add pagination element lol)
  - well see, probs add paging
- link to tournament viewer: link at client link ie "http://localhost:3000/?tournamentSource=https://api.battlecode.org/api/compete/bc25java/match/tournament/?external_id_private=private_bc25javasprint1_e746726c_dd4c_4c78_9fc8_1655c87ae2d4&format=json"

### DIFF
--- a/backend/siarnaq/api/compete/test_views.py
+++ b/backend/siarnaq/api/compete/test_views.py
@@ -406,6 +406,7 @@ class MatchSerializerTestCase(TestCase):
                     "maps": None,
                     "release_status": self.r_hidden.release_status,
                     "display_order": self.r_hidden.display_order,
+                    "in_progress": self.r_hidden.in_progress,
                 },
                 "participants": [
                     {
@@ -556,6 +557,7 @@ class MatchSerializerTestCase(TestCase):
                     "maps": [self.map.pk],
                     "release_status": self.r_results.release_status,
                     "display_order": self.r_results.display_order,
+                    "in_progress": self.r_results.in_progress,
                 },
                 "participants": [
                     {
@@ -635,6 +637,7 @@ class MatchSerializerTestCase(TestCase):
                     "maps": None,
                     "release_status": self.r_participants.release_status,
                     "display_order": self.r_participants.display_order,
+                    "in_progress": self.r_participants.in_progress,
                 },
                 "participants": [
                     {
@@ -714,6 +717,7 @@ class MatchSerializerTestCase(TestCase):
                     "maps": None,
                     "release_status": self.r_hidden.release_status,
                     "display_order": self.r_hidden.display_order,
+                    "in_progress": self.r_hidden.in_progress,
                 },
                 "participants": None,
                 "maps": None,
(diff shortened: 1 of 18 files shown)


nour-massri:
@lowtorola how is the progress going here? let me know if I can help

lowtorola:
This one is blocked by my questions here please feel free to take a look and work on this as much as you wish each out to Evan for mission clarifications

lowtorola:
- [ ] fix logic for create/enqueue
 - [ ] can do this when the round is active
 - [ ] active means TODO
- [ ] create "rerun round" functionality
 - [ ] function on "tournament round" model which requeues all of its matches
 - [ ] can rerun the round when it is active and none of its matches are in progress
- [ ] tidy up release round logic
 - [ ] can release when not released? have confirmation modal
- [ ] add link to challonge bracket
- [ ] add link to tournament viewer
- [ ] add pagination for tournament rounds
- [ ] debug why maps aren't actually being assigned to the round oopsie
- [ ] fix tournament round matches refetch (and all refetches) :)

nour-massri:
I went through most of it, I can say it looks nice with better error handling but I am not a big fan of changing user view if user is admin and might be more leaning towards adding the new actions and handle errors in the django admin interface not because it is nice but more like separation, 
also it would be cool if we can test with a non-public match currently we can't see them in tournaments and that's a reason I prefer more separation

lowtorola:
Strong disagree on doing tournaments through admin dashboard i think. @TheApplePieGod had more experience running tournaments but clearly it's just a giant PITA to do through the clunky admin interface. 

I do agree with preventing admin creep in play.battlecode though. I think it could be exclusively reserved for tasks like tpurney admin which are too complex to do through the simplistic Django admin field stuff. E.g. i think managing model fields etc should continue to be done via admin.api.battlecode which i assume you'd agree with 😉

Lmk what your thoughts are on these points and let's discuss further with the real stakeholders Evan Ruth and Nicole if were still on disagreement 😁

lowtorola:
I'm on board with that. I'm not remotely attached to the routing to the tournament admin page anyway either it was just the quickest solution lol

PR #907, opened 2025-01-14
Revert challonge v2.1

nour-massri:


n8kim1:
should be right (or at least, helpful)

PR #911, opened 2025-01-14
Rankings filter by eligibility ANDed

lowtorola:
- change filter to be an AND
- fix duplication bug (which i think was on the api...)


lowtorola:
sorry forgot to request from u nour lol

nour-massri (approved):

PR #912, opened 2025-01-14
Hide Unreleased Tournament Matches

lowtorola:
- removes all unreleased tournament matches from the match queryset unless the user is staff
- TODO: test!


nour-massri:
the python tests test the serializer filter, while we moved the filter to views,
also our views filter is missing filtering out tournaments that aren't public, is_public = false

nour-massri (approved):

Issue #908, opened 2025-01-14
explanative error for blocked ranked scrim


lowtorola:
we have the ability to turn off ranked scrims, but the FE shows the generic error message. it would be nice if we could display an error that tells the user why its being rejected.

PR #914, opened 2025-01-15
Returning the error message detail(when accepting/creating scrimmage) from backend to frontend 

nour-massri:


nour-massri:
closes https://github.com/battlecode/galaxy/issues/908

frontend/src/api/compete/useCompete.ts
```diff
@@ -375,17 +374,24 @@ export const useRequestScrimmage = (
...
+        } catch (e: unknown) {
+          const error = e as ResponseError;
+          // Parse the response text as JSON, detail propety contains the error message
+          const errorJson = (await error.response.json()) as {
+            detail?: string;
+          };
+          const errorDetail =
+            errorJson.detail ?? "An unexpected error occurred.";
+          throw new Error(errorDetail);
+        }
```
lowtorola, line 388: This is great! At some point we need to go through and make this a type-ified function which errors in detail for every frontend mutation.

lowtorola (approved):
Assuming youve tested this it looks great to me! see my comment about the refactored function, that would be a great issue to create :shipit:

PR #923, opened 2025-01-24
Fixes for Tourney Admin

lowtorola:
- altogether hides tournament matches for unreleased rounds, unless the user is staff or the matches are requested with external id private
- adds link to django admin for private challonge bracket :)
- fixes tournament match maps ordering with a lovely hack fix :sob: 
- dont hide round maps if the user is staff
- better error display in frontend when things go wrong (but not for initialize bc that is an asynchronous task)
- actually list maps in admin page


nour-massri (approved):
had to revert the no-staff filteration in serializler, need to debug this more

PR #952, opened 2025-11-28
add input validation for team eligibility changes

### LINKED ISSUE #939: No input validation in team eligibility allows adding internal flags like finalist
The api endpoint `/api/team/bc25java/t/me/` has very little input validation, allowing anyone to edit requests and set their eligibility flags to any of these values: 1, 2, 3, 4, 5, 6, 8, 9, 10.

Some of these are intentional, 1, 2, 3, 4, and 8 are accessible through the competitor UI, but 6 (previously finalists for 2024 and before), 5, 9, and 10 (not sure what those are, there are no teams with those flags in any year), are reserved. You can still set them, though.

## STEPS TO REPRODUCE

This is pretty easy to exploit using Firefox's "edit and resend" feature

1. Edit your team eligibility checkboxes and click save
2. Open the developer tools and go to "Network," then find the "PATCH" request to `/api/team/bc25java/t/me`
Image
3. Right click and click "edit and resend"
4. Change the "eligible_for" array to any combination of the above numbers
Image
5. Send the request

You can see that it works by editing the filter in the URL of the rankings page:

https://play.battlecode.org/bc25java/rankings?page=1&search=&eligibleFor=6

Image

The filter shows up as blank, and the teams are filtered correctly.

You can add yourself to the finalists list using this method (flag 9):

https://play.battlecode.org/bc25java/rankings?page=2&search=&eligibleFor=9

Image

The Realer Merlin, Pantheon, and SPAARK didn't compete in the finals but added themselves to the list through this method.

**One concern is that teams could add themselves to the finalists list, which probably would be caught, but may still mess with the bracket.** (I don't know if the brackets are automated, but it would cause issues)

nicolexxuu:
- Add new field` is_private` to `EligibilityCriterion` model + make db migrations
- Raise an exception if a competitor sends patch request to `/api/team/episode/t/me/` with criterion that is private or not assigned to episode `episode`
- Test erroring logic

Closes #939!

### DIFF
--- a/backend/siarnaq/api/compete/test_views.py
+++ b/backend/siarnaq/api/compete/test_views.py
@@ -1,7 +1,9 @@
 import io
 import random
+import re
 from collections import OrderedDict
-from datetime import timedelta
+from datetime import datetime, timedelta
+from datetime import timezone as dt_timezone
 from typing import Any
 from unittest.mock import mock_open, patch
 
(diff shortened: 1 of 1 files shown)


lowtorola (approved):
LGTM! great work!

PR #958, opened 2025-12-01
adding separate build triggers for staging

nour-massri:
This PR fixes this issue:
Currently cloud build has the same release tag regex that triggers build for both staging and production, 
We usually change it manually from cloud build to accept ^staging-siarnaq-* and ^staging-frontend-* to only deploy to staging when we want to test.


lowtorola (approved):
LGTM! :shipit:

PR #965, opened 2025-12-07
fix eligibility check bug

nicolexxuu:
- fix bug in eligibility check logic (prevent erroring when PATCH doesn't contain `eligible_for`


lowtorola (approved):
LGTM! :shipit:

